Issue #15, opened 2020-08-31
I want to access help for a standard AutoLISP function


ambrosl-adsk:
In the context of editing a LSP file, I want to select a function name and then right-click and choose a command like Help from the right-click menu.  This should open the latest version of the AutoLISP Reference documentation in a Web browser to the selected function's topic, if no text is selected it could just open Help to a general or Homepage page.

hansWurst-creator:
An additional option would be to include that information in the language server.

This extension shows what can easily be accomplished:
![image](https://user-images.githubusercontent.com/70693364/92107109-8b627280-ede5-11ea-97d5-f3a62e83817e.png)
[https://github.com/1833183060/autolisp](https://github.com/1833183060/autolisp)

A link to the corresponding help Article would be nice here as well.

JD-Howard:
> 
> 
> I want to select a function name and then right-click and choose a command like Help from the right-click menu. This should open the latest version of the AutoLISP Reference documentation in a Web browser to the selected function's topic, if no text is selected it could just open Help to a general or Homepage page.

Its still a little raw right now and I had to touch more things than I would of expected, but I do have this exact functionality working on my local copy. Just waiting to hear back on my _Individual Contributor Agreement_.

serhanbakir:
F1 key used to open up the online/offline help page about the selected function. Similar functionality will be very useful.

JD-Howard:
> 
> 
> F1 key used to open up the online/offline help page about the selected function. Similar functionality will be very useful.

Thank you for the feedback. That was my original intent too, but that simply won't happen because F1 is a very integral default key binding set to activating the All-Commands bar. The more you use VSCode the more you'll come to realize just how valuable that default binding is.

LanDong-adsk:
added "open online help" in V 1.4.0

PR #20, opened 2020-09-18
 Added right click 'Open Web Help' for Enhancement #15

JD-Howard:
Somebody (other than the new guy) really needs to go open the package.json with VSCode and save/commit/PR it. When I did that it lit up the entire file as changed because of refactoring and I sure didn't want to cause a reviewer a treasure hunt. I am not very familiar with the Git process to begin with so this seemingly inconsequential problem caused committing/reverting/Non VSCode editing/review/revert/retry steps I'd rather not do again.

Note, the 'webHelpAbstraction.json' file is expected to evolve into a better data backbone with some additional web scraping. I am certain there are some data type related information issues in there, but it should be absolutely fine to resolve the Web Help #15 feature request.


extension/src/completion/autocompletionProvider.ts
```diff
@@ -71,6 +86,7 @@ export function readAllBultinFunctions() {
             allCmdsAndSysvars.push(item)
     });
 
+    loadWebHelpAbstraction("../../extension/data/webHelpAbstraction.json");
```
hualin-adsk, line 89: Our extension can be used in Mac, I see some vla- prefix functions and objects, which are not supported in AutoCAD MAC.

Maybe this is not critical, we can left it there.

package.json
```diff
@@ -47,6 +47,11 @@
 				"category": "AutoLISP",
 				"title": "%autolispext.loadcurrentfile.title%"
 			},
+			{
+				"command": "autolisp.openWebHelp",
+				"category": "AutoLISP",
+				"title": "Open Web Help"
```
hualin-adsk, line 53: it needs to be put in the same file as above "%autolispext.loadcurrentfile.title%", which needs to be localized for different languages.

extension/src/completion/autocompletionProvider.ts
JD-Howard: I don't think this particular point is an issue. If I am working on a Mac, I've borrowed some code online and I trying to figure out why its breaking at a vla command, I would want the help to open that documentation so I could see it plainly stated it isn't supported in official documentation. Make sense?

package.json
JD-Howard: Okay, I see the package.nls.json that is the default and the various other language json files. Should be an interesting time with a translator to generate those.

My thought is to generate it once with all 3 words and convert them back. If they are 1-1, then they are fine, If they are not the same as the original input, then maybe try again with each individual word as input to find the disconnect.

JD-Howard: I believe I have all the language files (added 1 line to each) updated and the package.json using the pointer, but I ran into the same problem of VSCode refactoring the spacing of the indentations on json files. Is this some rogue extension I am using for basic HTML editing? Are any of these known to cause this problem?
abusaidm.html-snippets
alefragnani.Bookmarks
eamodio.gitlens
ecmel.vscode-html-css
eg2.tslint
eg2.vscode-npm-script
felixfbecker.php-debug
GitHub.vscode-pull-request-github
lonefy.vscode-JS-CSS-HTML-formatter
ms-mssql.mssql
ms-python.python
ms-vscode-remote.remote-containers
ms-vscode-remote.remote-ssh
ms-vscode-remote.remote-ssh-edit
ms-vscode-remote.remote-wsl
ms-vscode-remote.vscode-remote-extensionpack
mtxr.sqltools
mtxr.sqltools-driver-sqlite
pranaygp.vscode-css-peek
ritwickdey.LiveServer
sidthesloth.html5-boilerplate
sneezry.watch-in-chrome
Zignd.html-css-class-completion

JD-Howard: I found the JSON formatter option in the settings, I'll revert that commit and see if I that fixes the issue I've been having.

JD-Howard: For future reference this is what was causing the problem and why no settings I applied seemed to prevent json files from converting tabs to spaces on save.
**lonefy.vscode-JS-CSS-HTML-formatter**

I still have other problems regarding the lines I am editing, but at least it won't flag the whole file as a change now.

extension/src/completion/autocompletionProvider.ts
hualin-adsk: this is not a big issue, we can tell testers to accept this.

i18n/chs/package.i18n.json
```diff
@@ -1,6 +1,7 @@
 {
   "autolispext.description": "这是适用于 AutoCAD AutoLISP 的 vscode 扩展",
   "autolispext.loadcurrentfile.title": "加载文件到 AutoCAD",
+  "autolispext.openwebhelp.title": "打开在线帮助",
```
hualin-adsk, line 4: This is great!  I don't know how you translate these words, maybe google translator? 

In future maybe we can only change the i18n\enu\package.i18n.json file, then let the localization team in Autodesk to do left things.

hualin-adsk:
@JD-Howard Could you add more sections in this PR:
 ### abstraction 
   fill some big ideal, requirement, algorithm 

 ### screen shot 
 
 ### tests passed

i18n/chs/package.i18n.json
JD-Howard: Yes, this was all google translate. I converted it to the other other language and then back. I switched to using the word 'online' instead of 'web' because it was clearly more supported across the different languages. Luckily I only needed to concern myself with 3 words at this point.

I do think language localization is a highly specialized thing. If we have the option to outsource that to pros, then that would be the best course of action.

JD-Howard:
**Objective**
One of the features of the original VL IDE was the ability to highlight text and click the help button to pull up its documentation. My goal was to bring that functionality to the AutoLispExt.

**Implementation**
This was done by web scraping the online help to get all the unique GUID's that have identified these function in the web help since 2016. Assuming Autodesk maintains what they have been doing since 2016 moving forward, then the way I've designed this new feature should continue to work for the foreseeable future.

**Abstraction**
The added webHelpAbstraction.json file contains significantly more information than symbols and guids. It is the intent to further develop my scraping process to make this file suitable for representing function signatures and argument types. There is a desire to use this for context/index sensitive argument identifications while typing code, but that is probably a ways off and has a large degree of issues due the way signatures have been documented in the web documentation.

**Screenshots & Tests**
Here is an animated gif that should cover the screenshot & passed tests. If you need more than what I have provided here, then you'll have to be more specific on the expectations.
PR20

extension/src/project/projectCommands.ts
```diff
@@ -212,6 +213,16 @@ export function registerProjectCommands(context: vscode.ExtensionContext) {
             }
         }));
 
+        context.subscriptions.push(vscode.commands.registerCommand('autolisp.openWebHelp', () => {
```
hualin-adsk, line 216: I think this is not a command use by the "project: feature, it is only used in right click menu.
I suggest we move this function to other files, for example the same file for the "load Lisp" right click menu.

extension/src/completion/autocompletionProvider.ts
```diff
@@ -58,6 +60,19 @@ function readDataFileByLine(datafile: string, action: (items: string[]) => void)
         }
     });
 }
+
+// loads resources for the openWebHelp.ts objects.
```
hualin-adsk, line 64: we use this extension to format the TS codes:
https://marketplace.visualstudio.com/items?itemName=eternalphane.tsfmt-vscode

hualin-adsk:
thanks for your PR and detailed messages! This is a great feature!

extension/src/project/projectCommands.ts
JD-Howard: Should the openWebHelp.ts move out of the 'project' folder too, or just its registration?

Another reason I liked the notion of it being at the end of the call to registerProjectCommands was that its nearly the very last thing that loads. So even if somebody somewhere does experience problems with my addition, the majority of the extension would be fully loaded without error. I just want you to realize your telling me to move it ahead of a lot of things in the extension.ts.

image

hualin-adsk (approved):

JD-Howard:
I guess that localization step made me finally notice that my link constructor is specifically targeting /ENU/ and there are probably other language prefixes that need to be used instead. Being that I am not in those locations, I may need some help finding out what those are, but I'll certainly try to extrapolate.

extension/src/project/projectCommands.ts
hualin-adsk: I mean we need to move the whole openWebHelp.ts out of the 'project' folder. And I also suggest to move the register codes out of the projects commands (it is not project command).

maybe put your registration as step 6?

JD-Howard:
I can't truly envormentally test what a the vscode.env.language variable will spit out for users, but I tried to cover the basis and have tested the GUIDs with the standard domain and these return values to make sure they go where they are supposed to go.
image

extension/src/project/projectCommands.ts
JD-Howard: Okay, if you tell me the name of the folder the openWebHelp.ts should go in, I can do this. Should it be in its own folder called menus? I think that might make sense due to the other enhancement requests. How else could we "select a symbol and localize it within a defun"? IE, I suspect there will be more context menu commands coming....

JD-Howard:
Yeah I was just doing another commit anyway because I found this interesting bit in the VSCode 'bootstrap-window.js' while I was seeing if there was some kind of official list env.language could produce.
image

extension/src/project/projectCommands.ts
szilvaa-adsk: @JD-Howard I'd say you should create a `extension\src\help` folder and put all your source files there.

extension/src/project/openWebHelp.ts
```diff
@@ -0,0 +1,174 @@
...
+export function openWebHelp() {
+	const editor: vscode.TextEditor = vscode.window.activeTextEditor;
+	let selected: string = editor.document.getText(editor.selection);	
+	let urlPath: string = webHelpContainer.getWebHelpUrlBySymbolName(selected);
+	vscode.env.openExternal(vscode.Uri.parse(urlPath));
+}
+
+
+// This container object represents all the normalized data extracted from help.autodesk.com/view/OARX/
+export class WebHelpLibrary{
```
szilvaa-adsk, line 16: How does this extraction process work? Did you do it manually or do you have some automated process? I'm concerned about the maintenance effort. We (at least I) don't know how often these values (especially the guids) change.

szilvaa-adsk (changes requested):
Thank you for contributing!

extension/src/project/openWebHelp.ts
JD-Howard: So the standalone C# app that I have doing this work is a 2 step process that I currently have divided into 3 steps for efficiency and testing purposes. 

1. The left-side navigation system does actually have the links until they've been expanded and no amount of inline tweaking to the DOM exposes the link; IE, they must be clicked. So, first I just go expand all the things containing links I want to extract and export the HTML of that navigation system/delete anything I don't want as an extraction target. This is technically the only part of the process required to obtain guids because their links use them.

2. At this point the application (with minor tweaks) could go scrape from a URL or it can download the contents of the first <article> it finds to a local file. I've opted to download because it takes time to load pages before the DOM has all the required content. IE, the pages are more like templates and the server has to lookup what should be dynamically pushed into different parts of the template.

3. Scraping the document (regardless of source) is somewhat well structured. It doesn't have what I'd call perfect identifiers, but the first <h1> is **always** the title, the first <pre> is **always** the signature, etcetera. So, you can very easily identify sections and design rules around how those should be evaluated/stored in my custom object abstraction. Now, the content itself can get very subjective do to documentation inconsistencies and I don't know if I want to try explaining that in a comment, but that is why I have been saying I need to further develop the logic to extract good type data; which isn't currently being used, but structurally exists in the JSON object.

**GUID**
I am not worried at all about the guid system. Autodesk has maintained these since 2016 and across all languages. The only way these change is if they completely redesign their help system.

**Content**
Like the GUIDs, the actual contents of these help files hasn't really changed since 2016. The behaviors of these functions hasn't really changed since 2000! Granted there are some exceptions where commands were added, but those are far and few between; much more likely that if something is introduced it just won't be documented...

extension/src/project/openWebHelp.ts
```diff
@@ -0,0 +1,174 @@
...
+
+
+// Triggers VSCode to open the official web documentation URL related to the currently selected symbol
+// Note: this function is directly referenced by projectCommands.ts and the the package.json contributes (commands & menus) section.
+export function openWebHelp() {
+	const editor: vscode.TextEditor = vscode.window.activeTextEditor;
+	let selected: string = editor.document.getText(editor.selection);	
+	let urlPath: string = webHelpContainer.getWebHelpUrlBySymbolName(selected);
+	vscode.env.openExternal(vscode.Uri.parse(urlPath));
+}
```
szilvaa-adsk, line 12:
```suggestion
export async function openWebHelp() {
	const editor: vscode.TextEditor = vscode.window.activeTextEditor;
	let selected: string = editor.document.getText(editor.selection);
	if (selected === "") {
		await vscode.commands.executeCommand('editor.action.addSelectionToNextFindMatch');
		selected = editor.document.getText(editor.selection);
	}
	let urlPath: string = webHelpContainer.getWebHelpUrlBySymbolName(selected);
	vscode.env.openExternal(vscode.Uri.parse(urlPath));
}
```
This will allow right clicking on a word without having to select it.

extension/src/project/projectCommands.ts
```diff
@@ -212,6 +213,16 @@ export function registerProjectCommands(context: vscode.ExtensionContext) {
...
+            try {
+                openWebHelp();
+            }
+            catch (err) {
+                if (err)
+                    console.log(err.toString());
+            }
+        }));
```
szilvaa-adsk, line 224:
```suggestion
    context.subscriptions.push(vscode.commands.registerCommand('autolisp.openWebHelp', async () => {
            try {
                await openWebHelp();
            }
            catch (err) {
                if (err)
                    console.log(err.toString());
            }
        }));
```
Use async/await to ensure we can call async apis within `openWebHelp()`

szilvaa-adsk (commented):
Some suggestion to make the extension easier to use.

extension/src/project/openWebHelp.ts
szilvaa-adsk: We need to bring this C# app or its typescript/node equivalent into this repo so that anybody can run it and update the JSON.

JD-Howard: I completely agree and am committed to doing exactly that, but I need to do a **lot** of cleanup on it before I make it publicly available. It is safe to say it was an experiment that needs to be pull together in a fluid way and the above steps more formally documented first.

**Note**
I don't think I am remotely familiar enough ts/node to convert this in any kind of a timely manner. However, I do think that would be an excellent exercise for my own personal growth. If you want me to go that direction, then please don't hold up finalizing this PR over it.

extension/src/project/openWebHelp.ts
JD-Howard: I could combine this with a regex to identify the nearest contiguous string, but snippet you added doesn't actually provide a selection from an empty string starting point. However at that point I really wouldn't need to execute addSelectionToNextFindMatch anyway.

szilvaa-adsk: My goal was to allow the user to right click on a word choose help without having to select it. The 2 lines I added simply select the word under the cursor.

JD-Howard: I had a big long reply about how _addSelectionToNextFindMatch_ wasn't properly awaiting and that the text wasn't being selected until after the command finished, but decided to give it another try first and for some reason its working now.... Maybe I forgot to run the bat file or something; IDK, but its working!

I will say this is some interesting tribal knowledge because _addSelectionToNextFindMatch_ implies that you already have a selection.

JD-Howard:
- Updated and moved openWebHelp.ts to /src/help/
- Downloaded a fresh zip of the main and completely reverted all changes to ProjectCommands.ts & autocompletionProvider.ts
- Made legit edits to extension.ts to create my own entry points as specified.

That should be all the ToDo items other than making the abstraction creator available. When I do get the C# project cleaned up enough to put on Github, what do I do with it? Create a new public repo for it or I dump it in our new help folder? 

If it does go directly in this project, then I'll have to add a number of lines to the gitignore.

JD-Howard:
That last commit was just me aligning what was going on between the Lisp language.json and the DCL language.json files until I figured out why DCL comments were broken. Not much actually changed, but again I moved stuff around so I could compare them easier.

szilvaa-adsk (commented):
Can you please also move `webHelpAbstraction.json` into the help folder? There's no benefit in having it in a separate folder. It is integral part of your "help" solution. I understand that you wanted to follow the existing convention but I don't see any benefit in that.

JD-Howard:
I can, but I am not sure that is where it belongs. 
1. The data folder appears quite early in the project structure and is a completely generalized name.
2. I have to ask myself where the code related to contextual feedback from a mouse hover or while a user is typing will wind up. This is currently the most likely data source for anything we didn't dynamically extract (IE, user defuns) through evaluating the users code.

Lets say for a moment we get to the point of seeing a **setq** and decide we want attempt being aware of its potential data type by looking at the return type of its next function call. If that function type information is known, then we could allow the mouse hover on a symbol known as that variable (downstream from its declaration) to report that its "probably" a string, list or Layer object. At this point, that information would have to be based on this abstraction. So, I am asking if all things related to contextual feedback will also go in the help folder?

szilvaa-adsk:
> So, I am asking if all things related to contextual feedback will also go in the help folder?
I don't know. We will figure it out if/when such feature is implemented. For now, I think it is easier for everyone if everything related to the online help is in the help folder. If this data gets re-used for for other features then we can decide what to do.

szilvaa-adsk (approved):
I'd like to see approving reviews from @hualin-adsk  and @Sen-real  and then I'll merge this.

hualin-adsk (approved):

Sen-real (approved):
Cool feature! A few comments added.

extension/src/help/openWebHelp.ts
```diff
@@ -0,0 +1,212 @@
...
+//#endregion
+
+
+// Triggers VSCode to open the official web documentation URL related to the currently selected symbol
+// Note: this function is directly referenced by projectCommands.ts and the the package.json contributes (commands & menus) section.
+export async function openWebHelp() {
+	const editor: vscode.TextEditor = vscode.window.activeTextEditor;
+	let selected: string = editor.document.getText(editor.selection);	
+	if (selected === "") {
+		await vscode.commands.executeCommand('editor.action.addSelectionToNextFindMatch');
```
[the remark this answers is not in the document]
szilvaa-adsk: This selects word under the cursor. It seems to work for me.

extension/src/help/openWebHelp.ts
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	// Searches the library dictionaries for a reference to the provided symbol name. If found, yields help URL relevant to that symbol, but otherwise outputs generalized help URL
+	getWebHelpUrlBySymbolName(item: string): string {
+		let symbolProfile: string = item.toLowerCase();
+		let lang: string = this.getLanguageUrlDomain();
+		if (symbolProfile in this.functions){        
+			return this.functions[symbolProfile].getHelpLink(lang);
+		} else if (symbolProfile in this.enumerators){
+			return this.enumerators[symbolProfile].getHelpLink(lang);
```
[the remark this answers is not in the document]
JD-Howard: This is a fine example of inconsistencies in the Help documentation causing 1-off scenarios that requires special rules. My goal was to provide help documentation links for the many named enumerators that really only exist whenever a method/property has referenced them. Seeing this I probably should have made them be a minimum of 5 characters or something, but in general this appears to be working exactly as intended because of the way they did the layout of the page.

image

extension/src/help/openWebHelp.ts
```diff
@@ -0,0 +1,212 @@
...
+// Note: this function is directly referenced by projectCommands.ts and the the package.json contributes (commands & menus) section.
+export async function openWebHelp() {
+	const editor: vscode.TextEditor = vscode.window.activeTextEditor;
+	let selected: string = editor.document.getText(editor.selection);	
+	if (selected === "") {
+		await vscode.commands.executeCommand('editor.action.addSelectionToNextFindMatch');
+		selected = editor.document.getText(editor.selection);
+	}
+	let urlPath: string = webHelpContainer.getWebHelpUrlBySymbolName(selected);
+	vscode.env.openExternal(vscode.Uri.parse(urlPath));
```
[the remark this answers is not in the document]
JD-Howard: If the user asks for help on a subject we haven't index, we shouldn't them the help general landing page?

extension/src/help/openWebHelp.ts
```diff
@@ -0,0 +1,212 @@
...
+//#region Web Help Initialization
+// Invoked from extension.ts to setup the Right Click 'Open Online Help' command. 
+export function registerHelpCommands(context: vscode.ExtensionContext) {
+	context.subscriptions.push(vscode.commands.registerCommand('autolisp.openWebHelp', async () => {
+		try {
+			await openWebHelp();
+		}
+		catch (err) {
+			if (err) {
+				console.log(err.toString());
```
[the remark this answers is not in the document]
JD-Howard: do you want me to update the projectCommands.ts to export that function or do you want me to copy it to openWebHelp.ts?

extension/src/help/openWebHelp.ts
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	getDefaultHelpLink(languageDomain: string): string{
+		let year: number = new Date().getFullYear() + 1;
+		return "https://help.autodesk.com/view/OARX/" + year.toString() + languageDomain;
+	}
+
+	// Searches the library dictionaries for a reference to the provided symbol name. If found, yields help URL relevant to that symbol, but otherwise outputs generalized help URL
+	getWebHelpUrlBySymbolName(item: string): string {
+		let symbolProfile: string = item.toLowerCase();
```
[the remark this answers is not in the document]
JD-Howard: yep, that is a good idea to maximize the success rate

extension/src/help/openWebHelp.ts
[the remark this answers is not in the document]
JD-Howard: I had this exact same problem where *editor.action.addSelectionToNextFindMatch* wasn't actually selecting text until after my function had fully completed. After an updated call to *./winCompile.bat* it suddenly started working. 

I also understand your confusion because the name implies we already have a selection. If you search the page for the word "tribal" you'll find my comments on this with szilvaa-adsk.

extension/src/help/openWebHelp.ts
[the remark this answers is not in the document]
szilvaa-adsk: Please export the existing function and call it.

extension/src/help/openWebHelp.ts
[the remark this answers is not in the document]
Sen-real: Yes, opening the help general landing page is a great idea, I'm just a little bit worried that users then might not know it's supposed to select a word first, maybe you could update these files to describe your new feature:
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/main/README.md 
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/main/CHANGELOG.md
I can't upload image for some unknown problem, but when you select an extension in the "Extensions" panel which is by default on the left sidebar of VS Code, the "Details" and "ChangeLog" tabs will render above files.

Sen-real (approved):

extension/src/help/openWebHelp.ts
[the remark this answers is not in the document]
Sen-real: OK... I guess it's the cleanup step in the./winCompile.bat that helped. This batch file is supposed to do a clean build.

JD-Howard:
I can update those files, but you'll have to tell me what version number you want used in the change log.

Also, I can update the README.md, but it doesn't appear to be structured for documenting particular sub-feature. Seems more like the general features are just bullet points. Should I just add one that says "Right Click -> Open Online Help (selection sensitive)"

hualin-adsk:
I think the next release should be 1.4.0, one day the old VLIDE is really retired, then we can make the version to 2.0.0

Autodesk developer will do the clean up things in readme and changelog file before releasing it.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## [1.4.0] - 2020-09-21
```
hualin-adsk, line 1: This is heads up that the feature will be tested by the testers, then it can only be released to vscode market. So we will change the date in future!

Thanks for contributing this!

JD-Howard:
I leaned towards help.autodesk.com being the actual source of feature documentation and just gave the readme.md a bullet point. However, I did add an ./extension/src/help/openWebHelp.md for a more detailed explanation the testers can use.

Everything except the web scraper should be implemented that I am aware of...

**Note:** I tried to test this by using "gulp package" and manually installing the vsix, but the "webHelpAbstraction.json" did not make the trip. It worked fine after I put it in the right spot manually. I can't tell how to make that configuration change to the project.

hualin-adsk:
You mean that the webHelpAbstraction.json file is not packaged in the VSIX file via command "`gulp package`" ?

JD-Howard:
Correct, I had to copy it over manually and I couldn't see any special handling setup for the ./data/ folder to emulate.

hualin-adsk:
I think we have only file` .vscodeignore` to ignore packaging the files.

JD-Howard:
Well, the rule in there is a pretty broad scope that I doubt you want entirely gone. Can we do something like an "include" that will override that .vscodeignore for a single file?

hualin-adsk:
I don't think we have such a file, I think the ignore file name is hard code by the package tool.

Sen-real:
> Well, the rule in there is a pretty broad scope that I doubt you want entirely gone. Can we do something like an "include" that will override that .vscodeignore for a single file?

I think it is because the json is under the "extension/src" folder which we entire don't want to include in the extension package.

Below are some samples:
1) AutoLispExt\extension\data
files here are not excluded by the .vscodeignore
2) AutoLispExt_PM\AutoLispExt\pack.py
there're steps to copy a few files:
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/main/pack.py#L35
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/main/pack.py#L37

BTW, as the file name indicates, you can also make the package (autolispext.vsix) by "python pack.py" (it's python 3)

hualin-adsk:
@JD-Howard I think you can refine the .vscodeignore file, then it can be merged to main branch

JD-Howard:
I'll get that done, but may as well wait until I have time this weekend to finish cleaning up the abstraction creator. I've added more things to its extraction process like DCL Tiles/Attributes & whether its platform is "WIN", "MAC" or "WIN|MAC". So, I'll update the vscodeignore when I can dump the C# app in the Utils folder.

Also, contemplating whether a lot of the vla-get/set property functions should often wind up in ambiguous functions with different 1st argument signatures when they apply to many different object types. Already had to do that with Methods because they genuinely had some wildly different Inputs/Outputs. For properties, I was previously just populating a list of "validObjects" to capture that target info and was just representing that 1st argument super generalized to type: object. The biggest con to doing this is a fairly large size increase on the resulting JSON file because much of those new additions will repeat data. The pro is better context with an easier path to tie out to a specific situation while the user is typing and the ability to add/remove edge cases in post processing with a maintained set of rules when the documentation uses terms like "Read-Only: Yes Except for...". I don't think any of what I am describing affects the "Open Online Help" feature, but does have effects on the future utility of the abstraction.

I will have some other (hopefully minor) changes to expand the the support for the DCL document type.

JD-Howard:
Well, I wound up doing a lot more than just cleaning up the code. I pretty much normalized all the data and reviewed about 2/3 of the 2098 possibly-applicable help pages against the output and tweaked anything reasonable to peruse until I got what I was looking for. This included a complete reconstruction of all the syntax signatures. Here are some examples of old vs new:
<pre>
(+ [number number ...])

(+ number?&lt;integer|real>...)
</pre>
<pre>
(<= numstr [numstr ...])

(<= numstr&lt;many> numstr?&lt;many>...)
</pre>

<pre>
(vl-string-mismatch str1 str2 [pos1 pos2 ignore-case-p])

(vl-string-mismatch str1&lt;string> str2&lt;string> pos1?&lt;integer> pos2?&lt;integer> ignore-case-p?&lt;boolean>)
</pre>

This was not easy due to the number of 1-offs or inconsistencies that exist all over the help documentation. I definitely made the code less disgusting than it was, but all this special handling certainly prohibited me from getting the level of cleanliness I had hopped to achieve given the time available. I believe it was worth it though, I was able to perfectly capture like 95%-98% of what I wanted. However, I need more time to build out an "overrides" system for force feeding corrected values into things that are just too (special) much effort to be worth trying to auto-capture.

I've attached the latest version of the abstraction if anyone wants to look at it. I'll be jammed up until Wednesday, but I'll try to get this finished and pushed later this week.

**Regarding the above signature samples and assuming nobody objects to my new representations, is it okay if I throw in a mouse-hover tooltip on the next commit that shows these if available?** It is a very minimal amount of code to do this and would be a nice early warning indicator for whether a symbol has documentation.

[webHelpAbstraction.zip](https://github.com/Autodesk-AutoCAD/AutoLispExt/files/5293527/webHelpAbstraction.zip)

szilvaa-adsk:
> Regarding the above signature samples and assuming nobody objects to my new representations, is it okay if I throw in a mouse-hover tooltip on the next commit that shows these if available

I'm OK with that but I would prefer if we didn't grow this PR. Ideally, we should merge this and then you would create another PR for the mouse hover.

JD-Howard:
Okay, we'll isolate the PR to only include the C# app and adding the 'Open Online Help' to DCL file type.

utils/HelpAbstractionGenerator/ReadMe.md
```diff
@@ -0,0 +1,47 @@
+##Creating the webHelpAbstraction.json file
```
hualin-adsk, line 1: Thanks for adding this file!

hualin-adsk (approved):

utils/HelpAbstractionGenerator/ReadMe.md
JD-Howard: Well, it was a reasonable assumption for me (at this point) that it would be asked for... lol

I only finished the C#, I still have to update some stuff to hookup the help system for the DCL filetype.

Still struggling with trying to get the JSON file copied to the right location in the vsix. If you look at these 2 copy functions from the pack.py, I don't understand what this could accomplish. I followed the one that was already there, but mine looks like its targeting itself by doing that. I have tried putting it in ./out/src/help/ however that isn't the path that the TS is trying to read and the build or install process spawned a bunch of (i18 related) like-named jsons with it. This thing doesn't want to naturally copy over and VSCode is doing weird things with it. I don't suppose we could just call it a .txt and revisit dumping it somewhere that isn't going to fight me tooth and nail?

```python
def copyAcadProcFinder():
    src = os.path.join(os.path.curdir, 'utils', 'acadProcessFinder', 'bin', 'acadProcessFinder.exe')
    dst = os.path.join(os.path.curdir, 'out', 'process')
    copyFile(src, dst, 'copied acadProcessFinder.exe')

def copyWebHelpAbstraction():
    src = os.path.join(os.path.curdir, 'extension', 'src', 'help', 'webHelpAbstraction.json')
    dst = os.path.join(os.path.curdir, 'extension', 'src', 'help')
    copyFile(src, dst, 'copied webHelpAbstraction.txt')
```
I also took the src/** .vscodeignore and expanded it, but I think my src/* is probably still blocking everything in the folder. I had hoped that the single * would only block the immediate src files and let me manage the subdirectories.
```
src/*
src/help/openWebHelp.md
src/help/openWebHelp.ts
src/completion/**
src/format/**
src/process/**
src/project/**
src/test/**
```

utils/HelpAbstractionGenerator/StepBehaviors/Step4.WebObjectLibrary.cs
```diff
@@ -0,0 +1,1040 @@
...
+            HAP.HtmlNode retNode = template.returnValHeaders.First();
+            HAP.HtmlNode ulNode = retNode.getDoc().SelectSingleNode(retNode.ParentNode.XPath + "//ul");
+            string readonlyValue = retNode.nextRealSibling().InnerText.getLetters();
+            string typeValue = string.Join(", ", retNode.nextRealSibling().nextRealSibling().InnerText.Normalization().splitByAnyChar(" ,"));
+
+            var outVal = new ejoValueType("Return", typeValue, ulNode, linkCategory.Property);
+            var inVal = new ejoValueType("NewValue", typeValue, ulNode, linkCategory.Property);
+
+            bool isReadOnly = false;
+            if (readonlyValue.Contains("READONLYYES") && readonlyValue.Contains("EXCEPT") == false)
```
JD-Howard, line 912: Probably missing some vla-set functions that are valid simply because of the EXCEPT rule. Depending on how this gets used in the future, I probably need to create the setters, but somehow find the things they are excluding and also include them in our setters.

utils/HelpAbstractionGenerator/ReadMe.md
hualin-adsk: I am not sure if this works:` I had hoped that the single * would only block the immediate src files and let me manage the subdirectories.` If it can work, it is great. Otherwise I suggest to move the webHelpAbstraction.json to` Extension/Data` folder.

JD-Howard: Little surprised to see that TypeScript didn't have any native support for conditional build statements. So, I dirtied the function signature and built my own version of it. The code below lets us keep the file in /src/help/ and still generates a vsix with a working reference to the data file. If this is acceptable, shall I include my changes to Pack.py in the commit?

FYI, the added (i18) json files were for my ts->js not the copied file.

My *openWebHelp.ts*
```typescript
export function readAllHelpData() {
	loadWebHelpAbstraction(
		"../../extension/src/help/webHelpAbstraction.json",
		"../../out/help/webHelpAbstraction.json");
}
function loadWebHelpAbstraction(srcFile: string, outFile: string): void {
	webHelpContainer = new WebHelpLibrary();
	var fs = require("fs");
	var resolvedPath = path.resolve(__dirname, outFile);
	if (fs.existsSync(resolvedPath) === false) {
		resolvedPath = path.resolve(__dirname, srcFile);	
	}
    fs.readFile(resolvedPath, function(err, data) {        
        if (err === null) {                      
            webHelpContainer.load(JSON.parse(data));
        }
    });    
}
```
The *Pack.py*
```python
def copyWebHelpAbstraction():
    src = os.path.join(os.path.curdir, 'extension', 'src', 'help', 'webHelpAbstraction.json')
    dst = os.path.join(os.path.curdir, 'out', 'help', 'webHelpAbstraction.json')
    copyFile(src, dst, 'copied webHelpAbstraction.json')
```

hualin-adsk: Maybe this is an option: `come to think of it, I could also just add the same /out/help/ copy to the winCompile.bat and only reference that path in the TS. `

And please think about to put the webHelpAbstraction.json in `extension/data`, and other data files are also in that folder. I don't want to have a black sheep.

JD-Howard: I think the complications with not having it there totally validates what your saying, but this is just where I was told to put it https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/20#pullrequestreview-492208349
> Can you please also move `webHelpAbstraction.json` into the help folder? 

Either way, I'll put it wherever now that I know how to work around it.

hualin-adsk:
I agree that now the PR is so large, let's fix the json location issue, then merge it. 
Other changes should be created as a second PR.

JD-Howard:
Well, I think that is everything.... 

Thank you for baring with me through this. I will strive to not cluster this up in the future. 

Probably could have been a lot worse since I had zero experience with the VSCode API and very limited nodejs/electron/typescript/git experience.

winCompile.bat
```diff
@@ -4,3 +4,5 @@ call npm run compile
 
 echo copying acadProcessFinder.exe...
 copy utils\acadProcessFinder\bin\acadProcessFinder.exe out\process
+echo copying webHelpAbstraction.json...
+copy extension\src\help\webHelpAbstraction.json out\help
```
hualin-adsk, line 8: If this feature also wants to support on Mac, maybe we need to change the macCompile.bat

hualin-adsk (approved):

hualin-adsk:
@szilvaa-adsk I think now we can merge the PR.

PR #24, opened 2020-10-07
Added gotoProvider & Enhanced autoCompletion

JD-Howard:
I still have some further testing to do with this, but I'd rather get the review process going before I am too far down a rabbit hole. There are some things that could potentially be considered sketchy, but some effort was made to mitigate the potential problems too. Either way, I should really get the course correction now if its necessary.

**Note:** Do try to think about future benefits of using this to provide inline argument signatures for user defined functions.
- If a workspace is open, this will load every LSP into memory for quick referencing. I've tested this with 10MB of LSP's and it only bumps the memory footprint by less than 50mb. I figured on a CAD machine that's nothing and 10mb isn't realistic for most folks. What this does is make GoToDefinitionProvider (F12) nearly instantaneous across the whole workspace. All of this memory-document loading seems to be happening passively without any locking of VSCode.
- Adds workspace file management to account for all these memory-only documents. Some of this is in the form of vscode.workspace.events, but some had to come from a FileSystemWatcher because the workspace events only fire from direct input on the workspace interface.
- Enhanced the AutoCompletionProvider to use diverse symbols in its results to identify
   - Global Variables (contextual to current defun)
   - Argument Variables (contextual to current defun)
   - Localized Variables (contextual to current defun)
   - Functions (Documented; contextual to the ./data/ text files & the webHelpAbstraction.json)
   - Functions (Undocumented; contextual only to the ./data/ text files)
   - Function (User-Defined; contextual to all opened/previously activated LSP's)
   - Enumerations (Windows Only and probably won't keep it. This probably only makes sense when inline-function specific context is implemented for the webHelpAbstraction signatures and the enum suggestions can be narrowed down to what is possible)

### DIFF
--- a/extension/src/completion/autocompletionProvider.ts
+++ b/extension/src/completion/autocompletionProvider.ts
@@ -2,6 +2,12 @@ import * as vscode from 'vscode';
 import * as os from 'os';
 import * as path from 'path';
 import { isCursorInDoubleQuoteExpr } from "../format/autoIndent";
+import { LispParser } from '../format/parser';
+import { Sexpression, LispAtom } from '../format/sexpression';
+import { findContainers, ElementRange } from "../format/autoIndent";
+import { webHelpContainer } from "../help/openWebHelp";
+import { activeDocuments } from '../workspace/workspaceInit';
+import { ReadonlyDocument } from '../project/readOnlyDocument';
 
 let internalLispFuncs: Array<string> = [];
 let internalDclKeys: Array<string> = [];
@@ -115,6 +121,7 @@ function getCmdAndVarsCompletionCandidates(allCandiates: string[], word: string,
 
 function getCompletionCandidates(allCandiates: string[], word: string, userInputIsUpper: boolean): Array<vscode.CompletionItem> {
     let suggestions: Array<vscode.CompletionItem> = [];
+    let platform = (os.platform() === "win32") ? "WIN" : "MAC";
     allCandiates.forEach((item) => {
         var candidate = item;
         if (userInputIsUpper)
@@ -122,9 +129,18 @@ function getCompletionCandidates(allCandiates: string[], word: string, userInput
         else
             candidate = item.toLowerCase();
 
-        if (candidate.startsWith(word)) {
-            var label = candidate;
-            const completion = new vscode.CompletionItem(label);
+        if (candidate.startsWith(word) && (platform === "WIN" || !winOnlyListFuncPrefix.find(pref => candidate.startsWith(pref)))) {
+            let label = candidate;
+            let lower = label.toLowerCase();
+            let cKind: vscode.CompletionItemKind = AutoLispKind.vsKind(AutoLispKind.Symbol);
+            if (webHelpContainer.ambiguousFunctions[lower] && webHelpContainer.ambiguousFunctions[lower][0].platforms.includes(platform)){
+                cKind = AutoLispKind.vsKind(AutoLispKind.Function);
+            } else if (webHelpContainer.functions[lower] && webHelpContainer.functions[lower].platforms.includes(platform)){
+                cKind = AutoLispKind.vsKind(AutoLispKind.Function);
+            } else if (webHelpContainer.enumerators[lower]){
+                cKind = AutoLispKind.vsKind(AutoLispKind.Enum);
+            }
+            const completion = new vscode.CompletionItem(label, cKind);
             suggestions.push(completion);
         }
     });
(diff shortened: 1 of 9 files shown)


extension/src/completion/autocompletionProvider.ts
```diff
@@ -171,30 +187,222 @@ function getMatchingWord(document: vscode.TextDocument, position: vscode.Positio
...
-    else {
-        return allSuggestions.filter(function(suggestion) {
-            for (var prefix of winOnlyListFuncPrefix) {
-                if (suggestion.label.startsWith(prefix)) {
-                    return false;
+
+    private traverseSexpression(sx: Sexpression): void {
+        let context: number|string = "";
+        sx.atoms.forEach(atom =>{
+            if (context !== "EXIT" && atom instanceof Sexpression){
```
JD-Howard, line 246: just pointing out that this use of 'EXIT' instead of just doing a return has to do with it not actually exiting the traverseSexpression function.  Not quite sure why, if you have any ideas why return wouldn't (even after changing void to boolean) please let me know; I don't understand!

extension/src/completion/autocompletionProvider.ts
```diff
@@ -171,30 +187,222 @@ function getMatchingWord(document: vscode.TextDocument, position: vscode.Positio
...
+            }
+        });
+    }
+
+    load(document: vscode.TextDocument, position: vscode.Position): void {
+        // get the scope of this position  
+        let kwRegex: RegExp = /\(\s*(DEFUN|DEFUN-Q)\s+\S{1,99}(\s|\(|;)/gi;
+        let tmp = findContainers(document, position);
+        let pr1: ElementRange = tmp.containerBlockComment;
+        let pr2: ElementRange[] = tmp.containerParens;
```
JD-Howard, line 317: I meant to delete all this, was troubleshooting an error

extension/src/format/parser.ts
```diff
@@ -120,7 +120,7 @@ export class LispParser {
                     endPos2d = document.positionAt(endOfString.offsetInDocument);
                 }
                 else {
-                    endPos2d = selectionStartOffset + textString.length;
+                    endPos2d = selectionStartOffset + textString.length; // BUG: Doesn't produce Position Object
```
JD-Howard, line 123: If you could save me the time of deciphering all of this and just give me some idea of what this probably should be returning, I'll fix/test it. I only ever hit this bug only during extremely malformed requests. IE, the document was the constructor context, but I only asked it to tokenize a single (possibly nested) defun; made some epic bad decisions with those two very different context points. And why I wound up generating a ReadonlyDocument for my smaller scope of work.

extension/src/workspace/gotoProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+		return findGoToLocation(selected);
+	}
+}
+
+
+function findGoToLocation(searchFor: string): Promise<vscode.Location | vscode.Location[]> {
+	let kwRegex: RegExp = new RegExp("\\(\\s*(DEFUN|DEFUN-Q)\\s+" + utils.escapeRegExp(searchFor.toUpperCase()) + "(\\s|\\(|;)", "gi");
+	let found: Map<string, vscode.Position> = new Map();
+	// Try to find the definition in the currently opened documents
+	activeDocuments.forEach(doc =>{
```
hualin-adsk, line 28: I think only `activeDocuments` is not enough, please think about the following case:
1. user open a .prj file in vscode Autolisp side bar
2. click a file in .prj tree to open a lisp file
3. go to definition 
Maybe this definition is in other lisp file referenced by this .prj file, in fact we EXPECT the `goto` feature can work for this case.

extension/src/workspace/workspaceInit.ts
```diff
@@ -0,0 +1,175 @@
...
+import { AutolispGoToProvider } from "./gotoProvider";
+import { ReadonlyDocument } from "../project/readOnlyDocument";
+import * as fs from 'fs';
+
+
+// This variable represents un-opened *.LSP documents that we need to collect information from
+export const workspaceDocuments: Map<string, ReadonlyDocument> = new Map();
+export const activeDocuments: Map<string, vscode.TextDocument> = new Map();
+const docSelector = [ 'autolisp', 'lisp'];
+// How bad is it for me to store this context? I saw some other extensions doing this and vscode.extensions.getExtension() doesn't create a vscode.ExtensionContext
```
hualin-adsk, line 11: I suggest to create a global GotoContextManager class to hold all these global vars, including the above the `activeDocuments`

extension/src/completion/autocompletionProvider.ts
```diff
@@ -171,30 +187,222 @@ function getMatchingWord(document: vscode.TextDocument, position: vscode.Positio
...
+        } else {
+            // some kind of structural document error came into play
+            debugger;
+        }
+        this.previousItems.length = 0; // cleanup the tracking mess
+
+        // quick scan any open documents for defun headers. 
+        // Note: This workspace.textDocuments array only ever contains a single document or I would of used it instead.
+        
+        activeDocuments.forEach(doc =>{
```
hualin-adsk, line 351: I suggest to include all opened documents, and all the lisp files included in current .prj file

extension/src/completion/autocompletionProvider.ts
```diff
@@ -171,30 +187,222 @@ function getMatchingWord(document: vscode.TextDocument, position: vscode.Positio
...
-            for (var prefix of winOnlyListFuncPrefix) {
-                if (suggestion.label.startsWith(prefix)) {
-                    return false;
+
+    private traverseSexpression(sx: Sexpression): void {
+        let context: number|string = "";
+        sx.atoms.forEach(atom =>{
+            if (context !== "EXIT" && atom instanceof Sexpression){    
+                this.traverseSexpression(atom as Sexpression);
+                if (context === 2){
```
hualin-adsk, line 248: I don't understand this variable, what's the mean for 2 or 1?

extension/src/completion/autocompletionProvider.ts
hualin-adsk: This is recursive function, to return in the middle of recursion you need to set a flag, then check the flag for every recursion. To change the return value doesn't work.

extension/src/completion/autocompletionProvider.ts
```diff
@@ -171,30 +187,222 @@ function getMatchingWord(document: vscode.TextDocument, position: vscode.Positio
...
-function getLispAndDclCompletions(document: vscode.TextDocument, word: string, isupper: boolean): vscode.CompletionItem[] {
-    let currentLSPDoc = document.fileName;
-    let ext = currentLSPDoc.substring(currentLSPDoc.length - 4, currentLSPDoc.length).toUpperCase();
-    let candidatesItems = internalLispFuncs;
-    if (ext === ".DCL") {
-        candidatesItems = internalDclKeys;
+// Made a derived enum so we can swap out the vscode.CompletionItemKind values in a single place
+// Image of most symbols: https://camo.qiitausercontent.com/fb2a29c3301ab07f8af7cfa657752bdd9a62b8ea/68747470733a2f2f71696974612d696d6167652d73746f72652e73332e61702d6e6f727468656173742d312e616d617a6f6e6177732e636f6d2f302f3138393034312f66396263613830662d333136362d643338382d386538362d3632613535313363616431332e706e67
+export enum AutoLispKind {
+    Global = vscode.CompletionItemKind.Property,
```
hualin-adsk, line 193: Thanks for enhancing the AutoCompletion feature!

hualin-adsk:
Thanks for enhancing this!

I think we need to log two issues to track it (and testers latter can know how to verify it, those issue can only be closed after tester verified them, it means that maybe some bugs will be found).

And can you normalize the PR messages for the sections we mentioned in previous PR?
#### Objection
#### Implementation
#### Abstraction
#### Screen Shot & tests

extension/src/format/parser.ts
hualin-adsk: I think this position is the index to treat the multiline text as plain text. So this is a index for whole document.
@Sen-real can correct me at this point.

extension/src/completion/autocompletionProvider.ts
JD-Howard: That is fair and I considered making this an enum multiple times. If you don't understand then I certainly need to fix it. Essentially _Context_ is a dynamic state flag that talks in strings for DEFUN's and integers for SETQ's. If you think about a setq, it is a series of nested VarName (1) = Value (2).

extension/src/completion/autocompletionProvider.ts
JD-Howard: Do -> While it is....
Somewhat glad to hear this does make sense to you, but it is just weird to me because a _Return_ in C# would have broken out of whatever recursion level it was at.

extension/src/workspace/gotoProvider.ts
JD-Howard: Thanks for reminding me, I completely forgot about this context. Based on what you are saying, I think we can consider the PRJ references at the same precedence as truly Active documents. So, if you have no objection I'll lump those into the activeDocuments variable.

extension/src/workspace/workspaceInit.ts
JD-Howard: I was pretty worried about push back on this level of document management, but if I just need to be more formally handled, then I can do that.

extension/src/completion/autocompletionProvider.ts
JD-Howard: The only way I've found to get an accurate representation of what is truly open on-load is setting up the _workspace.onDidOpenTextDocument_ event and puppeteer the interface to cycle through all the available tabs. Which is **really** ugly, but our plugin isn't loaded until after the workspace has reloaded its previous state. 

This cycle-sync and I think a few other things about this update could probably use some new User-Options. I certainly wouldn't want this cycle to happen every time I hit F5, but I wouldn't mind it if it happened once at the start of my workday.
```typescript
const startingDocument = vscode.window.activeTextEditor.document.uri.fsPath;
	if (startingDocument.toUpperCase().slice(-4) === ".LSP"){
		do {
			await vscode.commands.executeCommand('workbench.action.nextEditor');
		} while (vscode.window.activeTextEditor.document.uri.fsPath !== startingDocument);
	}
```

JD-Howard:
> And can you normalize the PR messages for the sections we mentioned in previous PR?

Yes, but based on your responses and some of my own ideas for resolving these concepts, I probably need to go reconfigure quite a few things. One that occurred to me and you didn't mention is just moving the _CompletionSymbolCollector_ into its own *.TS file would probably clean up the existing code base changes. You also suggested that I go make a context manager (context.ts) and my instinct is to do the same with the file management stuff so it isn't mixed with the GoToProvider hook. I am somewhat wondering if I should collect a bit more feedback to put me on the right path, cancel this PR for now and revisit this when its all been largely restructured. I certainly don't want to create another chaotic mess!

extension/src/completion/autocompletionProvider.ts
hualin-adsk: ` our plugin isn't loaded until after the workspace has reloaded its previous state.`
I think our plugin is loaded when opening a LSP files, hit F5 key, see this [page ](https://code.visualstudio.com/api/references/activation-events)

extension/src/completion/autocompletionProvider.ts
```diff
@@ -171,30 +187,222 @@ function getMatchingWord(document: vscode.TextDocument, position: vscode.Positio
...
+        this.previousItems.length = 0; // cleanup the tracking mess
+
+        // quick scan any open documents for defun headers. 
+        // Note: This workspace.textDocuments array only ever contains a single document or I would of used it instead.
+        
+        activeDocuments.forEach(doc =>{
+            for (let lnum = 0; lnum < doc.lineCount; lnum++) {
+                const line: vscode.TextLine = doc.lineAt(lnum);
+                const value: string = line.text.toUpperCase();
+                if (value.search(kwRegex) !== -1){
```
hualin-adsk, line 355: I think this can not handle the` (defun)` if the line is in comment (specially for the block comment).
I think we need to use the **parser** to parse the whole file, the parser can tokenize the whole lisp file, and can find our interested atoms like (defun) (defun-q)

extension/src/completion/autocompletionProvider.ts
```diff
@@ -171,30 +187,222 @@ function getMatchingWord(document: vscode.TextDocument, position: vscode.Positio
...
+                break;
+            }
+        }
+        // Fallback for when the typed values are not in a Defun
+        if (contents === ""){
+            contents = document.getText(new vscode.Range(document.positionAt(containers.slice(-1)[0].startPos.offsetInDocument), document.positionAt(containers.slice(-1)[0].endPos.offsetInDocument)));
+        }
+        
+        
+        // Create and load that scope into a temporary document for contextual parsing, had to add this openFromContent method to narrow the LispParser SOW
```
hualin-adsk, line 335: I think we can parse the whole file, don't need to find the scope container firstly. Please see the class `LispParser`

hualin-adsk:
`One that occurred to me and you didn't mention is just moving the CompletionSymbolCollector into its own *.TS file would probably clean up the existing code base changes. `
Thanks! I think we need to make the PR to follow the [SRP rule](https://en.wikipedia.org/wiki/Single-responsibility_principle#:~:text=The%20single%2Dresponsibility%20principle%20(SRP,functionality%2C%20which%20it%20should%20encapsulate.). I see in this PR you covered two things: Auto-complete refinement; Goto definition feature. It is best to split the PR into two PRs.

extension/src/completion/autocompletionProvider.ts
JD-Howard: Thanks, I'll see if I can find something in that link to give me the document access, but right now our activate isn't called until after its too late. IE, If you close VSCode with a workspace active and 5 documents open, it will open back up with 5 documents tabs, but our extension has zero context about 4 of them until they are activated.

extension/src/completion/autocompletionProvider.ts
JD-Howard: One of the reasons I took this path is because of it being insanely fast. Can I add a LispParser variable to the ReadonlyDocument class so this will be pre-cached for me?

JD-Howard:
The SRP is tough for this because the AutoComplete work is contingent on the document handling. So, I think we are now saying the infrastructure PR has to happen and be merged before the AutoComplete PR can even be generated.

extension/src/completion/autocompletionProvider.ts
JD-Howard: I tried the all inclusive "*" to see if our Activate () would trigger soon enough to capture the workspace state reload and it didn't. Which kind of makes sense because [this page](https://vscode-docs.readthedocs.io/en/stable/extensionAPI/activation-events/) states that _Extensions are activated lazily in VS Code_.

I really don't like the notion of cycling the interface and that was actually the foundation of why I made that enhancement request for Import comment headers. If those existed, I really wouldn't care if the total scope of open documents were known/available because an explicit specification would ultimately be 10x better than including the defuns from some random document the user just happened to open, but may or may not be associated with something they care about in the active LSP.

hualin-adsk: Can we reference other open source extension like python for vscode? Or search some help from the vscode slack https://vscode-dev-community.slack.com/

JD-Howard: I've tried finding a public invite/join link to that slack community, but I never did find one.

I do think our problem is somewhat unique due to the archaic nature of AutoLisp. The vast majority of languages wouldn't bothered trying to solve this issue because they don't need it. They already have some version of 'import' that clearly defines what they should be looking at. AutoLisp is a globalized free for all and we don't really have that luxury.

I've been looking at [this](https://github.com/DonJayamanne/pythonVSCode/) python repo off and on for a while. They have some interesting stuff worth paying attention to, but nothing I related to the problem of finding documents that were opened prior to extension activation. At least nothing stands out to me that I haven't already tried.

JD-Howard: I don't know why I didn't think of this sooner. I went and searched the vscode repo for slack and found the link there

JD-Howard: That slack channel will be a big help, but unfortunately all the chatter suggests this is long standing issue that hasn't been addressed in VSCode yet. They have both of these namespaces returning single objects regardless of what is open and it seems like I am not the only one who doesn't understand why these things are named something that they simply are not.
vscode.workspace.textdocuments
vscode.window.visualtexteditors

hualin-adsk: Thanks for taking time to do researching! Maybe we leave some technical debt here, then do cleanup if in future find some methods to do it.

JD-Howard: Yep, that makes a lot more sense than implementing a deplorable workaround. I'll add some prominent TODO notes.

As soon as Sen-Real gets back to us on that one bug I'll probably be canceling this PR to go do a bunch of reconfiguration and scope it only to the context/document manager for now. Maybe this bug isn't a concern because it only fires on malformed (multi-line for sure) strings. Maybe we should just do a console log message, let it do what it can and fail where it does?

szilvaa-adsk:
The smaller the PR the easier it is to review and therefore faster to merge. Please split this up in 2-3 PRs that build on one another.

JD-Howard:
Will refactor and split this appart

extension/src/format/parser.ts
Sen-real: @hualin-adsk: I agree with @JD-Howard, it seems wrong. It's used to create a `vscode.Range `object, which should be created with either 2 "`vscode.Position`" or 4 "`Number`".
For Syntax, I think it should be `document.positionAt(selectionStartOffset + textString.length)`; but from logic point of view I can't tell if it produces expected result or not.

JD-Howard: I already know of a couple documents and/or bad requests that trigger this. So, I'll try the document.positionAt() and see if it returns something sensible.

Sen-real: Cool!

PR #26, opened 2020-10-11
Consolidated Resources

JD-Howard:
**Objective**
Consolidated the loading of resources into a single place and standardized the contract for loading JSON data objects.

**Implementation**
Created a new resources.ts file to house all data related IO operations and resulting memory stores. Removed data related IO from places previously handling this work and replaced their local variables with imported versions.

**Abstraction**
My long term goal here is to include a _get Data (){ return Resources; }_ in the upcoming context manager to make this information more freely available through a single standardized import.

**Tests Passed**
Verified the fields were in fact being loaded and that the previous coding using their own versions was still functional.

**Screenshots**
Not required


JD-Howard:
There was a bit of minor maintenance done to the openWebHelp.ts too, but all this was did was expose more features that already existed in the webHelpAbstraction.json file anyway.

hualin-adsk:
Thank you for doing this refactoring!

JD-Howard:
Doing multiple PR's is proving to be a pain too. New commits to my main updated this PR. Okay, fine, makes sense. However, creating a new branch and trying to create a PR from it wants to submit the changes from this PR too. Does this require multiple forks? This is going to take a while if I have to wait for each PR to merge...

**Planned PR's**
- Consolidate Resources (reference provided by Context Manager)
- Region Folding
- ReadonlyDocument Manager (reference provided by Context Manager)
- Context Manager
- AutoComplete Update
- Definition Provider

extension/src/help/openWebHelp.ts
```diff
@@ -133,7 +117,10 @@ enum WebHelpCategory {
 	PROPGETTER = 2,
 	PROPSETTER = 3,
 	FUNCTION = 4,
-	ENUM = 5
+	ENUM = 5,
+	DCLTILE = 6,
+	DCLATT = 7,
```
hualin-adsk, line 122: this is added by intent?

hualin-adsk (approved):

extension/src/help/openWebHelp.ts
JD-Howard: Yes, these put the TypeScript abstraction in sync with the C# versions.

hualin-adsk:
I didn't try this process if there are many PR in parallel:
 1. create a branch from your forked repo
 2. commit code to the branch 
 3. create a PR between your branch and the Autodesk-AutoCAD:main

JD-Howard:
Unfortunately, that process you describe appears to be including this PR's changed files too. There should only be 5 files and the only other thing the interface will really let me do that states 5 files is if I create the pull on my JD-Howard\AutoLispExt\ instead of the Autodesk-AutoCAD\AutoLispExt\
image

hualin-adsk:
Maybe we can try this:
 3. create another branch and commit codes
 4. create PR between the 2 branches in step 2 and 3
 5. After the first PR created in step2 is merged, then change the base for PR in step4

Maybe this is more complex, it is only for parallel PRs. (Maybe small PR can be merged very quickly.)

JD-Howard:
So far, the only thing I've consistently done with Git is expect "feature x" to do "stuff", but instead it does "things" that weren't desired and now require "more things" to roll back all the things I didn't want it to do in the first place; all while keeping a perfect record of how shitty my day has been. It seriously seems like trying to fix something often just creates an even deeper hole. So at this point, I generally just go create a local backup, do a hard reset on my repo, copy back what I need and move on when things go wrong. Beats wasting a bunch of time riding the git-go-round anyway.



I do think I might know how to solve this moving forward, but we have to get this PR committed before I am willing to try the various invasive things (partly described above) that are required to create a proper starting point to **exclusively** work on branches.

hualin-adsk:
@szilvaa-adsk please help to merge it

szilvaa-adsk (approved):

PR #27, opened 2020-10-13
Added Folding Configurations

JD-Howard:
**Objective**
Add user designated folding regions

**Implementation**
Updated *.JSON language & snippet files to facilitate code folding starting from _;#region_ (LSP) _//#region_ (DCL) identifiers and ending with _;#endregion_ (LSP) _//#endregion_ (DCL) closing markers.

**Abstraction**
Not Applicable

**Tests Passed**
I have verified that these tags are being honored in both LSP & DCL file. Even verified they are functioning in nested scenarios. Also tested the Snippet worked with a pre-selection by validating it would wrap around the currently selected text.

**Screenshots**
PR Folding


hualin-adsk:
Thanks for the new feature!

Do you think to add a right click menu for this feature?
And by default user can collapse the section via the matched () symbols, it still works well with this change?

JD-Howard:
FYI, I am splitting what was going to be a PR into 2 really tiny ones just so I can test the multiple PRs. It will be kind of important because I really need to post the document handler and the context manager as a pair or I'll wind up having to include revisions to the document handler with the context manager. Collectively the both of them aren't quite done because I am currently trying to figure out the best place to provide a reference to the current prjNode.

JD-Howard:
I agree, a right click menu option on the snippet trigger would be a nice option. I just read the other day that submenus have been introduced  so I would now be less reluctant to add context menu items knowing we can bury some if they get overwhelming. I'll look into this.

I don't follow your Match() question. In the image it is collapsing 2 defuns using the region markers not one.

hualin-adsk:
See this one pic:
image

At each left parenthesis you can click to collapse the section

JD-Howard:
I think your saying that we can already fold at pretty much every pair of parenthesis. That is a really nice feature that I do use, but regions are useful for grouping several functions into a single heading. For example, I have a really complex hanger routine that dynamically enhances 5 different types of hangers, but its UI is stupid simple for the users because my code has 5 different defun tests to determine which (if any) are valid. I would much rather collapse all 5 of those tests once than 5 individual defuns.

I will add the context menu item. I just need to go research dynamically invoking the snippet. I do understand that submenus are brand new and we don't currently have enough menu items to consider using them right now. I was just commenting that its nice to know we can (sometime in the future) nest seldom used ones if they start getting away from us. The work I've done on variable type decoding for autoCompletion makes me think I can hook up the right click -> localize variable feature fairly easily. So, we do have more context menu items coming and I wasn't real sure allocating one for regions was safe to do for menu space requirements.

hualin-adsk:
`its UI is stupid simple for the users because my code has 5 different defun tests to determine which (if any) are valid. `
Yes, it is simple and it is built in features from vscode. 
And we want to keep this simple block and your new feature block features both work well.

hualin-adsk (approved):

JD-Howard:
Very possible if you do want to add the right click menu version of this. I am just unclear where to put this random call to register a command other than the extension.ts
```typescript
context.subscriptions.push(vscode.commands.registerCommand('autolisp.addFoldingRegion', async () => {
	try {
		const snip = new vscode.SnippetString(';#region ${1:description}\n${TM_SELECTED_TEXT}\n;#endregion');
		await vscode.window.activeTextEditor.insertSnippet(snip);
	}
	catch (err) {
		if (err){
			let msg = localize("autolispext.commands.addFoldingRegion", "Failed to load snippet");
			showErrorMessage(msg, err);
		}
	}
}));

```

hualin-adsk:
Maybe add a ts file like blockRegion.ts to hold the command codes? Then reference it in extension.ts

hualin-adsk:
Thanks! How about commands.ts? Then all other short commands in future can be put in that file.

JD-Howard:
The _commands.ts_ is in there now. I went ahead and relocated the 1-off help loader to help establish a pattern in the file and I like how it cleaned up the header of the _openWebHelp.ts_ file. You'll have to decide (later) if you want to migrate the other existing (non-project) command subscriptions over. These are represented by Steps 1 & 4 in the _extension.ts_ so they could represent a critical and/or sequenced infrastructure that cannot be moved.

I also updated the main _package.nls.json_ file and used that pointer in _package.json_, but I believe you told me someone better suited would handle the individual language ones. Is that still correct or would you like those update those with whatever google translate spits out?

extension/src/commands.ts
```diff
@@ -0,0 +1,40 @@
...
+			}
+		}
+	}));
+	
+
+	// Associated with the right click "Insert Region" menu item
+	context.subscriptions.push(vscode.commands.registerCommand("autolisp.insertFoldingRegion", async () => {
+		try {
+			let commentChar = vscode.window.activeTextEditor.document.fileName.toUpperCase().slice(-4) === ".DCL" ? "//" : ";";
+			const snip = new vscode.SnippetString(commentChar + "#region ${1:description}\n${TM_SELECTED_TEXT}\n" + commentChar + "#endregion");
```
hualin-adsk, line 28: The region can be nested? or what will happen if user create nested region?

hualin-adsk (approved):

extension/src/commands.ts
JD-Howard: Regions do nest, I can add a GIF of that if you'd like

hualin-adsk:
It is ok for localization at this point. We will ask the professional localization team to do translate. And it is a process we release the extension.

extension/src/commands.ts
hualin-adsk: ok, I only remind us that there is a such case, and I think tester will also cover this case.

hualin-adsk (approved):

hualin-adsk:
@szilvaa-adsk I think this can be merged if there is no other comments coming.

extension/src/commands.ts
Sen-real: is it able to trigger this command when there's no active text at all? or what if the active text is not lsp or dcl?

JD-Howard: Yes, this does trigger with no selection. It would just create an empty start/stop region block.

I believe the "when" property on the _package.json_ defines this for the "autolisp" and "autolispdcl" file selectors. So, if you were to open a *.TXT file, this won't even show up as an option. I tested this on a *._LS" file and the context menu option was missing.

Sen-real: > Yes, this does trigger with no selection. It would just create an empty start/stop region block.
> 
> I believe the "when" property on the _package.json_ defines this for the "autolisp" and "autolispdcl" file selectors. So, if you were to open a *.TXT file, this won't even show up as an option. I tested this on a *._LS" file and the context menu option was missing.

The "when" is inside and for a context menu item definition, but as shown in your GIF, as long as the lisp extension has been activated, the command can also be directly executed via. the Command Palette of VS Code.
Please have a test:
1. Open a LSP to activate the extension
2. Open a txt file
3. Open the Command Palette and execute the new added command

szilvaa-adsk (approved):

PR #28, opened 2020-10-13
Infrastructure Updates 

JD-Howard:
**Objective**
A small collection of infrastructure updates to enable upcoming enhancements. All of these should be inert at the moment, but they are small and concise so they should be easy to review for issues. Also gave me a chance to test a multiple PR workflow; which appears to be working.

The ReadonlyDocument was given embedded token information so that several (future) downstream processes would not be redundantly repeating the code required to perform this task on workspace/prj document references. Also suggesting we migrate an existing call on createProject() to a more reusable and ambiguous function.

**Tests Passed**
Used the existing findeContainers() and the new getRange() addition to test the changes to the ReadonlyDocument.

**Screenshots**
Not Applicable


extension/src/project/readOnlyDocument.ts
```diff
@@ -71,13 +73,21 @@ export class ReadonlyDocument implements vscode.TextDocument {
...
+    //Note to reviewers, this method is used exactly one place. We could update that call to provide 'autolispprj' to the more ambiguous createMemoryDocument()
     static createProject(prjContent: string): ReadonlyDocument {
         let ret = new ReadonlyDocument('');
 
         ret.initialize(prjContent, 'autolispprj');
         return ret;
     }
 
+    // used to create micro memory documents from fragments of AutoLisp code. One application for doing this is clearly identifying the scope of variables.
+    static createMemoryDocument(fileContent: string, languageId: string): ReadonlyDocument {
```
hualin-adsk, line 85: @Sen-real Please help to review it

extension/src/format/autoIndent.ts
```diff
@@ -5,13 +5,17 @@ import * as format from './listreader';
...
-class ElementRange {
+export class ElementRange {
     constructor() {
         this.startPos = null;
         this.endPos = null;
         this.quoted = false;
     }
 
+    getRange(document: vscode.TextDocument): vscode.Range{
+        return new vscode.Range(document.positionAt(this.startPos.offsetInDocument), document.positionAt(this.endPos.offsetInDocument));
```
Sen-real, line 16: If I remember it correctly, the end of range should be the index of the next char that is right after the range.
See https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/90a25dbe7e7282ff29346d5d22608070236db111/extension/src/format/autoIndent.ts#L39,
Please do some test to determine what to do.

JD-Howard: Thanks for the insight. I will look into this, it is entirely possible the ending parenthesis was missing and I just didn't notice.

extension/src/project/readOnlyDocument.ts
Sen-real: it looks good, basically an upgrade of https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/28/files#diff-b33bf13e4dee64627ec3c1c8acfddf97ca2cbf70824a82fae89605816c69e94bL74

JD-Howard: Thanks, I will track down that one call and update it to provide its selector

extension/src/format/autoIndent.ts
JD-Howard: Still working on creating a proper test for this and the multi-line string bug. In both cases I need to re-install some of the stuff I shelved in a backup to do it and make sure none of that winds up in a commit.

extension/src/project/readOnlyDocument.ts
```diff
@@ -97,12 +97,35 @@ export class ReadonlyDocument implements vscode.TextDocument {
...
         }
-        this.atomsForest = [];
-        if (languageId === 'autolisp'){
+    }
+
+    
+    // Converted this from a constant data feature into an on-demand feature that once used is essentially cached for future queries.
+    get atomsForest(): Array<string|Sexpression> {
+        if (this.languageId === 'autolisp') {
+            if (this._atomsForest && this._atomsForest.length > 0){
```
Sen-real, line 106: Is this field initialized as null or undefined? Considering that the user is possible to open an empty LSP file, what's the length of this array in that case? if it's greater than 0, I'm then OK with the second condition, otherwise I think we shouldn't check the length here, because it will update the cache for every call.

extension/src/project/readOnlyDocument.ts
```diff
@@ -97,12 +97,35 @@ export class ReadonlyDocument implements vscode.TextDocument {
...
+                this.updateAtomsForest();
+                return this._atomsForest;
+            }
+        } else {
+            return [];
+        }
+    }
+    
+    // This was segregated from the atomsForest getter to support two primary use cases:
+    //      A force update that will be called on the workspace.onDidSaveTextDocument() saved event to keep the memory document in sync with the user input.
```
Sen-real, line 118: Shall we consider this situation? If a file is opened, edited, and saved on VS Code, we should access it via. the ordinary TextDocument of VS.Code, not this ReadonlyDocument.
The use of this ReadonlyDocument is limited to read the current content from a file, and that's all. The reason why we have this class is because there're some VSCode APIs which are designed to work with a document, not with a string.

extension/src/project/readOnlyDocument.ts
```diff
@@ -82,6 +82,22 @@ export class ReadonlyDocument implements vscode.TextDocument {
...
     }
 
+    // Added this to essentially cast standard TextDocument's to a ROD type enabling work to be done with enhanced standardized/functionality
+    // Related to to discussion issue#17
+    static getMemoryDocument(doc: vscode.TextDocument): ReadonlyDocument {        
+        let ret = new ReadonlyDocument('');
+        ret.eol = vscode.EndOfLine.CRLF;
+        ret.eolLength = 2;
+        ret.lineCount = doc.lineCount;
+        ret.languageId = doc.fileName.toUpperCase().slice(-4) === ".LSP" ? "autolisp" : "";
```
Sen-real, line 92: Despite of the real purpose of this method, the post fix could be .PRJ,  .DCL, etc.
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/main/package.json#L281
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/main/package.json#L292
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/main/package.json#L301

Sen-real (approved):

JD-Howard:
This kind of makes me think we need to add a general purpose _detectSelectorType(fspath: string)_ function to the _util.ts_

There are several places in the ReadonlyDocument that could benefit from it, but I can see places all over the code base (using Search in Files) that could also be clean up by this. Maybe I will add this as a static method in the DocumentManager class; kind of makes sense to put it there...

JD-Howard:
Is there anything outstanding on this one? I've made some suggestions on how to clean things up, but they are dependent on the document/context manager PR so I will have to double back to resolve the one item I am aware of later. I will likely do a PR solely for updating many tiny parts of the code to use the Context/Doc Manager immediately after it looks like it will get approved.

My only other conversations in here had a discussion issue #30 generated to hopefully clarify that aspect.

extension/src/project/readOnlyDocument.ts
```diff
@@ -222,6 +276,37 @@ export class ReadonlyDocument implements vscode.TextDocument {
...
-
     //#endregion
+
+    // This is very similar to a DOM querySelector and is to be used for quickly finding all the Defun's or Setq's to determine available function/variable names.
+    // The 'all' variable determines whether the function will continue digging inside a function it was able to match for nested versions.
+    // Test Case: Temporarily added the following to extension.ts	
+    //              let rod = ReadonlyDocument.getMemoryDocument(vscode.window.activeTextEditor.document);
+    //              let forest = rod.atomsForest;
+    //              let expl = rod.findExpressions(/(DEFUN|LAMBDA|FOREACH)/ig, true);
+    findExpressions(regx: RegExp, all: boolean = false): Sexpression[]{
```
hualin-adsk, line 287: It is best that if we can add some tests for these functions.

hualin-adsk (approved):

hualin-adsk:
@szilvaa-adsk please help to merge it

szilvaa-adsk (approved):

PR #29, opened 2020-10-14
PRJ Treeview, ProjectNode & LspFileNode Updates

JD-Howard:
**Objective/Abstraction**
The primary goal was to add ReadonlyDocument references to the PRJ LispFileNode's for the upcoming document management system. I expanded this to a Project Treeview feature update that allows users to close open projects. I did this because the future definitionProvider is intended to search PRJ's and open documents first when applicable. If it can't find what it was looking for there, then it will search the entire workspace. This new behavior means that both opening and closing a PRJ's to set a different baseline context can be a valuable behavioral tool for the users. A good example of this would be if (many do) users have "archive" folders and they don't want false positives from them, then they can simply dump all their core LSP's into a PRJ. If they are wanting to drudge through their history, they can easily do that too by closing the active PRJ.

**Implementation**
Added a close command to the PRJ Treeview's context menu and button bar. You'll notice that the icons on the Treeview button bar now support dynamic visibility contextual to the presence of a PRJ. Everything user-facing should have been localized. PRJ's now automatically create their ReadonlyDocument reference to be used for searching. The upcoming document manager will have a _get projectDocuments()_ that finds/returns a vscode.TextDocument[] from those LspFileNode's if available.

**Tests Passed/Screenshots**
PR 29 PRJ UpdatesSmall

### DIFF
--- a/extension/src/project/projectCommands.ts
+++ b/extension/src/project/projectCommands.ts
@@ -1,4 +1,4 @@
-import * as vscode from 'vscode'
+import * as vscode from 'vscode';
 import { OpenProject } from './openProject';
 import { ProjectTreeProvider } from './projectTree';
 import { openLspFile } from './openLspFile';
@@ -62,6 +62,22 @@ export function registerProjectCommands(context: vscode.ExtensionContext) {
                     let msg = localize("autolispext.project.commands.openprojectfailed", "Failed to open the specified project.");
                     showErrorMessage(msg, err);
                 });
+        }));
+
+        context.subscriptions.push(vscode.commands.registerCommand('autolisp.closeProject', async () => {
+            try {
+                let promptmsg = localize("autolispext.project.commands.closepromptmsg", "Confirm close request on: ");
+                let responseYes = localize("autolispext.project.commands.closepromptyes", "Yes");
+                let responseNo = localize("autolispext.project.commands.closepromptno", "No");
+                vscode.window.showWarningMessage(promptmsg + ProjectTreeProvider.instance().projectNode.projectName, responseYes, responseNo).then(result => {
+                    if (result === responseYes){
+                        ProjectTreeProvider.closeProject();
+                    }
+                });
+            } catch (err) {
+                let msg = localize("autolispext.project.commands.closefailed", "Failed to close project.");
+                showErrorMessage(msg, err);
+            }
         }));
 
         context.subscriptions.push(vscode.commands.registerCommand('autolisp.addFile2Project', async () => {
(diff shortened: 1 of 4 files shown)


extension/src/project/projectCommands.ts
```diff
@@ -1,4 +1,4 @@
-import * as vscode from 'vscode'
+import * as vscode from 'vscode';
```
hualin-adsk, line 1: I like this refinement. I am curious which tool prompts you this point.

package.json
```diff
@@ -75,6 +75,15 @@
 					"light": "images/light/Open_Project.svg"
 				}
 			},
+			{
+				"command": "autolisp.closeProject",
+				"category": "AutoLISP",
+				"title": "%autolispext.project.closeproject.title%",
+				"icon": {
+					"dark": "images/dark/close.svg",
```
hualin-adsk, line 83: I am not sure if we need a professional designer to give us two icons, or Microsoft has some standard icons.

hualin-adsk:
@Sen-real Please help to review this PR.

extension/src/project/projectTree.ts
```diff
@@ -124,18 +124,25 @@ export class ProjectTreeProvider implements vscode.TreeDataProvider<DisplayNode>
 
     public updateData(newRootNode: ProjectNode) {
         this.rootNode = newRootNode;
-
+        vscode.commands.executeCommand("setContext", "autolisp.hasProject", ProjectTreeProvider.hasProjectOpened());
         this.onChanged.fire();
     }
 
     public refreshData(data?: DisplayNode) {
+        vscode.commands.executeCommand("setContext", "autolisp.hasProject", ProjectTreeProvider.hasProjectOpened());
```
Sen-real, line 132: This method is also used to refresh a single LSP file node. I think it's enough to just do it in the "updateData" method, in which we truely change the rootNode field.

JD-Howard:
Now that a project can be closed, are you open to the idea of hiding the New/Open project buttons while one is open. I noticed those two options weren't in the project right click and I am always a fan of minimizing button clutter.

Sen-real:
> Now that a project can be closed, are you open to the idea of hiding the New/Open project buttons while one is open. I noticed those two options weren't in the project right click and I am always a fan of minimizing button clutter.

Well when a solution is open in Visual Studio, I can open another one. Visual Studio will automatically close the first. 
Anyway if we want to add this feature, I think we can do it in a different PR.

Sen-real (approved):

package.json
hualin-adsk: @Sen-real do you have some comments on these icons?

Sen-real: @hualin-adsk I'm the one who uploaded these two icons

hualin-adsk (approved):

hualin-adsk:
@szilvaa-adsk please help to merge this PR.

PR #33, opened 2020-10-25
Style Guide

JD-Howard:
**Objective**
Standardize the way TypeScript should _fundamentally_ exist within AutoLispExt

**Additional Context**
I took the minimalist style guide example and played around with it. I am sure you'll want to change a bunch of it, but it is a starting point. I originally had a bunch of ==highlights== in the document to direct attention to notably questionable things, but apparently GitHub doesn't support the highlighter markup.

### DIFF
--- a/DevREADME.md
+++ b/DevREADME.md
@@ -111,15 +111,15 @@ Below are some general
 5. ==Use `I` as a prefix for interface names that define function declarations.==
 5. Use whole words in names where possible.
 
-#### Types
+### Types
 1. Do not export types/functions unless you need to share them across multiple components.
 2. Within a file, type definitions should come first.
 
-#### Comments
+### Comments
 1. ==Use JSDoc-style comments for functions, interfaces, enums, and classes.==
 2. ==Limited to using an informal description and @type/@param/@return when applicable==
 
-## Style
+### General
 1. Use arrow functions over anonymous function expressions.
 2. Always surround loop and conditional bodies with curly braces.
 3. Open curly braces always go on the same line as whatever necessitates them.
@@ -136,7 +136,7 @@ Below are some general
   let qux = 'These aren\'t the droids you\'re looking for.';  // escaped quotes in string
 ```
 
-## Classes
+### Classes
 1. Avoid public properties - use setters/getters.
 2. Do not use the `public` keyword for public members - it is implied by default.
 3. Initialize all private properties to an initial value or `null`, unless they 
@@ -151,7 +151,7 @@ Below are some general
    properties go last.
 6. Abstract methods go in the same location as they would in the implemented class.
 
-## Examples
+### Examples
 - Interface declaration:
 
 ```typescript
--- a/DevREADME.md
+++ b/DevREADME.md
@@ -101,6 +101,8 @@ If you encounter this problem, maybe your NODE is too old, in version v10.13.0 i
 can work well.
 or run "npm cache clean -force"
 
+---
+
 ## Style Guide
 Below are some general code guidelines that should be followed when adding or updating the code base. It is the expectation that participating members actively apply this style guide to legacy code during maintenance so we can eventually have a noticeably consistent developer experience.
 ### Names
--- a/DevREADME.md
+++ b/DevREADME.md
@@ -104,11 +104,11 @@ or run "npm cache clean -force"
 ## Style Guide
 Below are some general code guidelines that should be followed when adding or updating the code base. It is the expectation that participating members actively apply this style guide to legacy code during maintenance so we can eventually have a noticeably consistent developer experience.
 ### Names
-1. ==Use `PascalCase` for type/interface names, enums, setters, getters and global singletons==
+1. Use `PascalCase` for type/interface names, enums, setters, getters and global singletons
 2. Use `camelCase` for function, argument and local variable names.
 3. Use `UPPER_CASE` for global constants.
-4. ==Use `_` as a prefix for private variables==
-5. ==Use `I` as a prefix for interface names that define function declarations.==
+4. Use `_` as a prefix for private variables
+5. Use `I` as a prefix for interface names that define function declarations.
 5. Use whole words in names where possible.
 
 ### Types
@@ -116,20 +116,20 @@ Below are some general code guidelines that should be followed when adding or up
 2. Within a file, type definitions should come first.
 
 ### Comments
-1. ==Use JSDoc-style comments for functions, interfaces, enums, and classes.==
-2. ==Limited to using an informal description and @type/@param/@return when applicable==
+1. Use JSDoc-style comments for functions, interfaces, enums, and classes.
+2. Limited to using an informal description and @type/@param/@return when applicable
 
 ### General
 1. Use arrow functions over anonymous function expressions.
 2. Always surround loop and conditional bodies with curly braces.
 3. Open curly braces always go on the same line as whatever necessitates them.
 4. `else` goes on the same line as the closing curly brace.
-5. ==Use two (2) spaces for indentation.==
-6. ==The `export` keyword should be on its own line.==
-7. ==Function should always provide type declarations for its return and argument values==
-8. ==Function bodies (including getters/setters) must appear on separate lines from their signature declarations regardless of their simplicity. Empty constructors that exist for clarity are acceptable on a single line==
-9. ==Use a maximum of 140 chars per lines; including leading whitespace==
-10. ==Strings should always use single quotes if they do not contain interpolated values:==
+5. Use two (2) spaces for indentation.
+6. The `export` keyword should be on its own line.
+7. Function should always provide type declarations for its return and argument values
+8. Function bodies (including getters/setters) must appear on separate lines from their signature declarations regardless of their simplicity. Empty constructors that exist for clarity are acceptable on a single line
+9. Use a maximum of 140 chars per lines; including leading whitespace
+10. Strings should always use single quotes if they do not contain interpolated values:
 ```typescript
   let foo = 'bar';       // static string
   let baz = `${5 * 5}`;  // interpolated string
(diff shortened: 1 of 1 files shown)


DevREADME.md
```diff
@@ -99,4 +99,132 @@ You may encounter the failure of npm install, the error message would say:
...
+1. Avoid public properties - use setters/getters.
+2. Do not use the `public` keyword for public members - it is implied by default.
+3. Initialize all private properties to an initial value or `null`, unless they 
+   will be initialized in the constructor. If a basic type is used, do not add 
+   a type specifier (e.g. `private _someValue = 0;`).
+4. Ordering of members within a class should be:
+  - `static`
+  - `public`
+  - `protected`
+  - `private`
```
JD-Howard, line 152: This was the original order from the example document. Personally I'd like to see private variables directly under the class declaration.

DevREADME.md
```diff
@@ -99,4 +99,132 @@ You may encounter the failure of npm install, the error message would say:
...
+1. Use arrow functions over anonymous function expressions.
+2. Always surround loop and conditional bodies with curly braces.
+3. Open curly braces always go on the same line as whatever necessitates them.
+4. `else` goes on the same line as the closing curly brace.
+5. Use two (2) spaces for indentation.
+6. The `export` keyword should be on its own line.
+7. TypeScript errors/warnings need to be resolved.
+8. Function should always provide type declarations for its return and argument values
+9. Function bodies (including getters/setters) must appear on separate lines from their signature declarations regardless of their simplicity. Empty constructors that exist for clarity are acceptable on a single line
+10. Use a maximum of 140 chars per lines; including leading whitespace
```
JD-Howard, line 134: The original minimalist example value was 80 except for function declarations. I grabbed the value from the in-depth version because 80 was a bit to short.; IE, roughly half the width of a 1280x1024 monitor at what I believe is the default text size.

DevREADME.md
```diff
@@ -99,4 +99,132 @@ You may encounter the failure of npm install, the error message would say:
...
+1. Use JSDoc-style comments for functions, interfaces, enums, and classes.
+2. Limited to using an informal description and @type/@param/@return when applicable
+
+### General
+1. Use arrow functions over anonymous function expressions.
+2. Always surround loop and conditional bodies with curly braces.
+3. Open curly braces always go on the same line as whatever necessitates them.
+4. `else` goes on the same line as the closing curly brace.
+5. Use two (2) spaces for indentation.
+6. The `export` keyword should be on its own line.
```
JD-Howard, line 130: I don't personally like this, but I do think it would make exported items a lot more visible

DevREADME.md
```diff
@@ -99,4 +99,132 @@ You may encounter the failure of npm install, the error message would say:
...
+5. Use `I` as a prefix for interface names that define function declarations.
+5. Use whole words in names where possible.
+
+### Types
+1. Do not export types/functions unless you need to share them across multiple components.
+2. Within a file, type definitions should come first.
+
+### Comments
+1. Use JSDoc-style comments for functions, interfaces, enums, and classes.
+2. Limited to using an informal description and @type/@param/@return when applicable
```
JD-Howard, line 122: I updated the examples to actually reflect this

DevREADME.md
```diff
@@ -99,4 +99,132 @@ You may encounter the failure of npm install, the error message would say:
...
+
+### General
```
JD-Howard, line 124: - Add item to prefer "let" over "var" 
- Add item related to const. However, I don't really know exactly what to say about const. I've personally started using it to declare just about everything to purposefully cause errors and make me rethink what/why/how I am doing something.

DevREADME.md
```diff
@@ -99,4 +99,132 @@ You may encounter the failure of npm install, the error message would say:
...
+1. Use arrow functions over anonymous function expressions.
+2. Always surround loop and conditional bodies with curly braces.
+3. Open curly braces always go on the same line as whatever necessitates them.
+4. `else` goes on the same line as the closing curly brace.
+5. Use two (2) spaces for indentation.
```
JD-Howard, line 129: From what I can tell, we are standardized around a single tab, is that accurate?

DevREADME.md
```diff
@@ -99,4 +99,132 @@ You may encounter the failure of npm install, the error message would say:
...
+---
+
+## Style Guide
+Below are some general code guidelines that should be followed when adding or updating the code base. It is the expectation that participating members actively apply this style guide to legacy code during maintenance so we can eventually have a noticeably consistent developer experience.
+### Names
+1. Use `PascalCase` for type/interface names, enums, setters, getters and global singletons
+2. Use `camelCase` for function, argument and local variable names.
+3. Use `UPPER_CASE` for global constants.
+4. Use `_` as a prefix for private variables
+5. Use `I` as a prefix for interface names that define function declarations.
```
JD-Howard, line 113: I do realize that "I" is a C concept that is highly frowned upon, but I would really like a way to tell the difference between an interface object that is essentially a value type from one that is expected to provide functionality

DevREADME.md
```diff
@@ -99,4 +99,132 @@ You may encounter the failure of npm install, the error message would say:
...
\ No newline at end of file
+or run "npm cache clean -force"
+
+---
+
+## Style Guide
+Below are some general code guidelines that should be followed when adding or updating the code base. It is the expectation that participating members actively apply this style guide to legacy code during maintenance so we can eventually have a noticeably consistent developer experience.
+### Names
+1. Use `PascalCase` for type/interface names, enums, setters, getters and global singletons
+2. Use `camelCase` for function, argument and local variable names.
```
JD-Howard, line 110: Might be a good idea these into 3 statements:
- Use `PascalCase` for type/interface names, enums, and global singleton instance variables
- Use `PascalCase` for getters and setters within singleton classes
- Use `cabelCase` for functions, getters, setters, arguments and local variables

DevREADME.md
```diff
@@ -99,4 +99,132 @@ You may encounter the failure of npm install, the error message would say:
...
+10. Use a maximum of 140 chars per lines; including leading whitespace
+11. Strings should always use single quotes if they do not contain interpolated values:
+```typescript
+  let foo = 'bar';       // static string
+  let baz = `${5 * 5}`;  // interpolated string
+  let qux = 'These aren\'t the droids you\'re looking for.';  // escaped quotes in string
+```
+
+### Classes
+1. Avoid public properties - use setters/getters.
```
JD-Howard, line 143: I think this should read:
Avoid public variables in singleton classes and prefer private variables with managed getters/setters

DevREADME.md
```diff
@@ -99,4 +99,132 @@ You may encounter the failure of npm install, the error message would say:
 Error installing vscode.d.ts: Error: read ECONNRESET
 If you encounter this problem, maybe your NODE is too old, in version v10.13.0 it 
 can work well.
-or run "npm cache clean -force"
\ No newline at end of file
+or run "npm cache clean -force"
+
+---
```
hualin-adsk, line 104: it only needs to paste the github link to this file? Or we can adjust the guide lines. I think anyway we should paste the github link for the copyright

DevREADME.md
hualin-adsk: which item is for `let` and for `const`?

DevREADME.md
JD-Howard: I don't recall seeing a copyright on the minimalist version. Plus I did alter it quite a bit already with the intent for us to alter it a great deal more.

hualin-adsk:
@Sen-real @szilvaa-adsk please help to review this PR.

DevREADME.md
JD-Howard: The first bullet is standalone, but we could just say both items would read
_Use const or let instead of var_

I was just remarking that a clearer definition of when to use const might be helpful because I do find that it has been helping me work evolve my plans

DevREADME.md
JD-Howard: I went back through their wiki pages (all 2'ish of them) and there are absolutely no copyrights posted within their wiki. Their code base is licensed under "Revised BSD" and each of their actual *.TS files do in fact contain a copyright. 
- Maybe they also borrowed that wiki page conent as a starting point from somewhere and thus can't put their copyright on it? 
- Maybe this type of documentation (especially this abstract) isn't worth putting a copyright on? 

I am clearly not familiar with licenses in general, but this content seems very standard to the vast majority and highly subjective personal preference. I would be surprised if this winds up even 25% equal to the original by the time we are done injecting our own ideas and restructuring things.

DevREADME.md
```diff
@@ -99,4 +99,132 @@ You may encounter the failure of npm install, the error message would say:
...
+   * @returns {void}
+   */
+  clearFlag(flag: WidgetFlag): void {
+    this._wflags &= ~flag;
+  }
+
+  private _wflags = 0;
+}
+```
+
```
hualin-adsk, line 230: I want to added a sections for `Tools`:
```
 format tools: tsfmt
 Lint tool: tslint
```

JD-Howard: That is a great idea! I definitively had some problems with my extension set when I started. Do you think we should do _Required Tools_, _Suggested Tools_ & _Conflicting Tools_. Each is itemized in an unordered list with a very short description is sufficient or shall we go beyond that with important Settings information?
I think my question clarification would effectively swap the information on either side of your colon and then file them under Required Tools heading.

hualin-adsk: I personally use this tool to format the TS codes: https://github.com/vvakame/typescript-formatter, maybe there is other excellent formatting tools.
Only provide the `Suggested Tools` is enough?

szilvaa-adsk (changes requested):
I don't find Style Guides very useful unless they are enforced by tooling. So instead of spelling out all these rules in  English I'd rather see some tooling hooked into the build that enforces whatever style we want to follow.

szilvaa-adsk:
> Fair point, but don't we need to have a blueprint of what we are trying accomplish before we can possibly go configure tooling that enforces it?

I think we should simply choose the tool and configure it. We can discuss individual configuration items then. 

I think tslint is now [deprecated](https://github.com/palantir/tslint) so we should probably use eslint instead. We could also look at https://prettier.io/ to take care of our code formatting needs.

JD-Howard:
I will try some options and make a issue for discussing this migration to different tooling.

PR #37, opened 2020-10-26
Rollout of AutoLispExt.localize() & General import header maintenance

JD-Howard:
**Objective**
Use a "common" instance of `nls.localize()` across the entire code base per the specifications of the official _vscode-nls_ project documentation.

Import the AutoLispExt constant, remove all imports of vscode-nls and its config() variable. Update the remaining code to point to AutoLispExt.localize(). This did require some adjustment of the _extension.ts_ header to create the AutoLispExt singleton instance before our other /src/ imports tried referencing it. IE, the entire extension was failing to load before I did that; I added a comment to identify the criticality of its arrangement.

**Additional Context**
While I was doing this, I also did:
- performed some general maintenance on all the headers within files referencing vscode-nls to add semi-colons and replace `requires()` with `import` references
- fixed a function name in the utils.ts that misspelled 'active' so you will find the occasional instance where that name changed. 
- added a dev dependency to @types/fs-extra since some of our code was using it as a require() and converting it to an import caused complaints.

**Tests Performed**
I installed the Spanish language pack and the localization's I already know we have are working. However, I did notice that _F1 -> AutoLisp: Save Project_ without having one opened, did produced an error that was not localized. I don't believe we have one because nothing in the project appears to be configured to provide a localization of "_autolispext.project.saveproject.noprojecttosave_"

**Screenshots**
Not Applicable

### DIFF
--- a/extension/src/commands.ts
+++ b/extension/src/commands.ts
@@ -1,9 +1,8 @@
 import * as vscode from "vscode";
-import * as nls from 'vscode-nls';
+import { AutoLispExt } from './extension';
 import { openWebHelp } from './help/openWebHelp';
 import { showErrorMessage } from './project/projectCommands';
 
-const localize = nls.config({ messageFormat: nls.MessageFormat.file })();
 
 export function registerCommands(context: vscode.ExtensionContext){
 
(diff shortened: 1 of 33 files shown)


hualin-adsk (approved):

hualin-adsk:
@szilvaa-adsk I think it can be merged now

szilvaa-adsk (approved):

Issue #36, opened 2020-10-26
Cannot open project file


morbius1st:
**Describe the bug**
Open an existing project fails with the message that the project file is malformed.

**To Reproduce**
1.  create a project
2. add files to project
3. close project (open something else)
4. goto AutoLisp project sidebar
5. chose open existng project
6. select the project just made above
7. project does not open and dialog box is displayed that reads "failed to open the specified project.  Error: Malformed project file ..."

**Expected behavior**
Project file is opened and project list is updated

**Screenshots**
none

**Desktop (please complete the following information):**
 - OS: windows
 - Version 10.0.10941

**Additional context**
none

JD-Howard:
Jeff, 
Unfortunately that error is a somewhat generalized response to a failure of parsing as a whole and doesn't tell me much about why. If you attach the *.PRJ file your having issues with, I'd be happy to run it through the debugger and narrow down the why for you.

hualin-adsk:
And I also can not repro the issue to follow your steps!

morbius1st:
A PRJ file is attached.

I assume that you followed the procedure noted and it worked - yes?

Attached is a pic of the error message when I try to open the file.

[Start System.zip](https://github.com/Autodesk-AutoCAD/AutoLispExt/files/5443083/Start.System.zip)
![image](https://user-images.githubusercontent.com/17919240/97259637-82d05c00-17d8-11eb-8be2-f46677c362b5.png)

JD-Howard:
Thank you Jeff, I see the problem. 
It looks like PRJ names are not compatible with spaces, but our extensions PRJ interface lets you create them with spaces. To fix that file you'll have to change "Start System" to "StartSystem" and open up the file in a text editor to put StartSystem together on one line too.

morbius1st:
That fixed it - thanks.  Should be an easy fix in the code.

PR #41, opened 2020-10-27
Resolved Issue #36

JD-Howard:
**Objective**
Aligned our PRJ filenames with the VLIDE legacy rules. Issue #36 illuminated that users could create PRJ files containing spaces and that did not match classic representations of PRJ names or their resulting internal structures.

Added checks to see if the user provided path.basename() that contained spaces and threw a (localized) error to inform them of the issue.

**Additional Context**
Everything in these 2 files were manually aligned to PR #37 to avoid later merge conflicts. I'd probably still merge 37 before this one since we know this one adds on top of it.

**Tests Performed**
Created PRJ with and without spaces to see what succeeded/failed. Opened PRJ's with and without spaces to see what succeeded/failed.

### DIFF
--- a/extension/src/project/createProject.ts
+++ b/extension/src/project/createProject.ts
@@ -18,7 +18,7 @@ export async function getNewProjectFilePath() {
 
     let fileUri = await vscode.window.showSaveDialog(options);    
     if (fileUri) {
-        if (path.basename(fileUri.fsPath).split(' ').length === 1){
+        if (path.basename(fileUri.fsPath).indexOf(' ') >= 0) {
             return Promise.resolve(fileUri);
         } else {
             let msg = AutoLispExt.localize("autolispext.project.createproject.nospaces", "Legacy PRJ naming rules do not allow spaces");
--- a/extension/src/project/createProject.ts
+++ b/extension/src/project/createProject.ts
@@ -18,7 +18,7 @@ export async function getNewProjectFilePath() {
 
     let fileUri = await vscode.window.showSaveDialog(options);    
     if (fileUri) {
-        if (path.basename(fileUri.fsPath).indexOf(' ') >= 0) {
+        if (path.basename(fileUri.fsPath).indexOf(' ') === -1) {
             return Promise.resolve(fileUri);
         } else {
             let msg = AutoLispExt.localize("autolispext.project.createproject.nospaces", "Legacy PRJ naming rules do not allow spaces");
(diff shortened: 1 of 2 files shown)


extension/src/project/createProject.ts
```diff
@@ -19,17 +16,23 @@ export async function getNewProjectFilePath() {
...
-    if (fileUri)
-        return Promise.resolve(fileUri);
-
-    return Promise.resolve(undefined);
+    let fileUri = await vscode.window.showSaveDialog(options);    
+    if (fileUri) {
+        if (path.basename(fileUri.fsPath).indexOf(' ') === -1) {
+            return Promise.resolve(fileUri);
+        } else {
+            let msg = AutoLispExt.localize("autolispext.project.createproject.nospaces", "Legacy PRJ naming rules do not allow spaces");
```
hualin-adsk, line 24: not sure the "`nospaces`" is in AutoLispExt/i18n/enu/out

hualin-adsk:
@Sen-real Please help to review

hualin-adsk (approved):

extension/src/project/createProject.ts
JD-Howard: I perceived these arguments to localize to be Key & EnglishTranslation. Every time I create a new piece of user-facing text then it needed a new ID; Is that incorrect?
My assumption was the Autodesk Localization team would be using some kind of a tool like vscode-nl-dev to collect and create translations for these and all I really needed to do was make sure I use the localize() to give them that opportunity.

hualin-adsk: They have no tools, and it only translate all the json style strings in the folder `AutoLispExt/i18n/enu`. So for developer we need to put a key-value json pair in that folder.

JD-Howard: That sounds like an extensive treasure hunt across the code base. IE, I believe we have many Keys in the code that do not existing json entries. I've certainly added to them by not understanding localize wasn't creating harvesting opportunities.
I will either be locating those tools or creating them if I necessary, but I am certain some kind of diffs tool is out there already.

hualin-adsk: Some strings in the current repro are not handled by` localize()`, because developer think that this is impossible to have such case, so it doesn't do localize. I think there is only one case with `TODO ` which need to add to json file.

It is best if there is a tool to extract it and put it to json file.

Sen-real (approved):

Issue #31, opened 2020-10-19
Update Project to TypeScript 3.7


JD-Howard:
**Is your feature request related to a problem? Please describe.**
We are missing a major bit of functionality (null chaining) that would clean up our logical expressions.

The following throws an error in TS when trying to transpile to JS:
```typescript
if (vscode.window.activeTextEditor?.document)
````

**Describe the solution you'd like**
Update to TypeScript 3.7.0 or greater. This null chaining is [native to JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and TypeScript finally got around to supporting it. Our conditionals would really benefit (more concise) from this update.

**Additional context**
Here is the release notes about chaining.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html

I played around with doing this, but I think NodeJs (and maybe more) also needs to be updated in order to use 3.7

LanDong-adsk:
updated in v 1.4.0

PR #42, opened 2020-10-31
Setup Ci

szilvaa-adsk:
We use Github Actions to build every PR and we also create a release for every push into `main`. The release contains snapshot of the code + the vsix. The vsix is versioned as prerelease. We use `npm version` to create a new prerelease version number during the build  in `package.json` (we don't commit this change).

@hualin-adsk @Sen-real I haven't hooked up tests since these require VSCode. We may want to think about how to do that.


hualin-adsk:
It is cool, but I can not access this release page https://github.com/Autodesk-AutoCAD/AutoLispExt/releases/tag/v1.4.1-92611c1bfad8d6bf50a4c390bfee6944bb7ab083.0

And can we adjust to the version to 1.3.X? I think V1.4.0 is our next release version.

szilvaa-adsk:
Yeah, I'm still experimenting. I deleted the above release. That's why you don't have access.

package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "autolispext",
 	"displayName": "AutoCAD AutoLISP Extension",
 	"description": "This is a vscode extension for AutoCAD AutoLISP",
-	"version": "1.4.0",
+	"version": "1.3.0",
```
szilvaa-adsk, line 5: 1.3.0 is the last official release. `npm version` increments the patch version and appends the preid we specify (commit sha in our case). This results in version like 1.3.1-<sha> which is exactly what we want.

package.json
```diff
@@ -599,34 +599,35 @@
 		"test": "npm run compile && node ./out/test/runTest.js"
 	},
 	"devDependencies": {
```
szilvaa-adsk, line 601: I updated some of the packages.

.github/workflows/main.yml
```diff
@@ -0,0 +1,68 @@
...
+        id: update_version
+        # See https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions
+        run: |
+          $version=npm version prerelease --no-git-tag-version --preid=${{github.sha}}
+          $version=$version.substring(1)
+          echo "::set-output name=VERSION::$version"
+        shell: pwsh
+
+      - name: Install dependencies
+        run:  npm install
```
szilvaa-adsk, line 38: I don't bother to user yarn.

package.json
JD-Howard: Super excited about the one to TypeScript, thank you Albert!

Your doing some pretty heavy foundational stuff. So, I am going to locally backup the stuff I've been working on, wait for the dust to settle on this PR and start from a fresh main branch before I author another.

JD-Howard: Not necessary, but the "@types/chai" would be a nice to have. Now that everything else has rolled onto newer versions I don't suspect I'll have the same issues temporarily putting it in there locally though.

package.json
```diff
@@ -599,34 +599,35 @@
...
-		"tslint": "5.11.0",
-		"typescript": "3.2.4",
-		"vscode-test": "^1.3.0",
 		"mocha": "^6.1.4",
-		"chai": "^4.2.0",
+		"run-sequence": "^2.2.1",
 		"source-map-resolve": "0.6.0",
-		"vscode-nls-dev": "^3.2.6"
+		"tslint": "5.11.0",
+		"typescript": "^3.9.7",
```
hualin-adsk, line 618: So this will resolve the issue https://github.com/Autodesk-AutoCAD/AutoLispExt/issues/31

extension/src/project/projectTree.ts
```diff
@@ -123,7 +123,7 @@ export class ProjectTreeProvider implements vscode.TreeDataProvider<DisplayNode>
     public updateData(newRootNode: ProjectNode) {
         this.rootNode = newRootNode;
         vscode.commands.executeCommand("setContext", "autolisp.hasProject", ProjectTreeProvider.hasProjectOpened());
-        this.onChanged.fire();
+        this.onChanged.fire(this.rootNode);
```
hualin-adsk, line 126: I think this is the fix for complaining from higher typescript

.github/workflows/main.yml
hualin-adsk: I think white source tool needs the yarn to install.

hualin-adsk (approved):

extension/src/project/projectTree.ts
JD-Howard: Also want to share that these higher versions of TypeScript apparently changed getters/setters. From what I am reading it mostly affects inherited classes trying to use the base class getters/setters so I don't believe our code will be affected.

package.json
szilvaa-adsk: @JD-Howard Don't hesitate to add this. I don't want to muddy the PR with this unrelated change.

.github/workflows/main.yml
szilvaa-adsk: I added `CodeQL` security scanning to our CI. We can add others if we want.

hualin-adsk:
This is very useful! It at least to ensure that there is no build issue.

PR #43, opened 2020-11-04
Added i18 Utility & Updating Localizations

JD-Howard:
**Objective**
This application was put together to detect and help resolve `AutoLispExt.localize()` TypeScript references. This allows a division of concerns through focusing on code and not missing things that need to be JSON localized later. This also helps normalize all the files structures to a standardized (newtonsoft) representation. In the below JSON changes, this more often than not was purely a structural spacing change.

Secondary Objective: Identify and Resolve the outstanding localization issues with my prior PR's that I didn't understand how far into the process I was responsible for. **Currently uncertain** if I am actually supposed to go fill out anything other than the **ENU** references.

Refer to the [README.md](https://github.com/JD-Howard/AutoLispExt/blob/LangUtility/utils/autolispext.i18.helper/README.md) file for an itemized breakdown of what this dialog does. This C# .Net Core application isn't pretty, but it is doing the job it was designed to do. Also note, that I originally started making another screen to (optionally) remove keys from JSON files that were no longer referenced. When I didn't find any, I stopped working on that and just commented out what I had done.

**Tests Performed**
The JSON files in this PR are the results of opening performing no edits and just hitting save. So, the GIF below is not part of the PR result. I would like this to be looked at without the extra noise first and I can go back and do the google translate operations for each of the missing items on the next commit; **assuming that is what is desired.**

**Screenshots**
i18utility


hualin-adsk:
Hi Joshua,

This is a cool tool. 
I think this tool is only used to developer who want to use the extension which is currently in developing.
Before official release we have Localization team to do translation and test. So for our extension developer it only needs to fill the json file in **ENU** folder.

JD-Howard:
That last update caused an occasional bug where it displayed duplicate items. It didn't affect the output, but needed a constraint to match the number of languages.

Also added the ENU versions of everything that was missing. Note, this does still create json files (with empty values) for the other languages. Should just be an exercise of "fill in the blanks" for the localization team.

i18n/chs/out/commands.i18n.json
```diff
@@ -0,0 +1,4 @@
+{
+  "autolispext.help.commands.openWebHelp": "",
```
hualin-adsk, line 2: it looks like this is empty for `chs`

JD-Howard: Well, now maybe you can see why I am generally very confused about what I am supposed to do beyond ENU.

The **package.nls.th-tw.json** should have an entry for the context menu itself from a prior PR. I am guessing the localization your looking at now relates to the command palette and was apparently not generated in that prior PR.  Note, this utility made no attempt at managing the package.nls files since those did not match the structure and were not tied to *.ts files.

With all that said, this question only validates my uncertainty. Do you actually want me to go populate all of the missing language values for _autolispext.help.commands.openWebHelp_?

hualin-adsk: `Do you actually want me to go populate all of the missing language values for autolispext.help.commands.openWebHelp?`
I think it is related to our objections. If we want to support the localized string when developing, it needs to populate for all languages. Instead if we don't want to support the localized version, and before the official release there is professional localization team to translate all the string in the `ENU `folder and test them.

I suggest we only add the strings in the `ENU `folder and `package.nls.json`, then we have more time to do other enhancement.

JD-Howard: Then I would say the current state of this probably represents a "normal" developer state because it does have all the ENU strings populated. The "other" languages do have their keys loaded into the JSON files, but the values are just empty strings. Which should make it really easy for the localization team to identify what they need to do work on.

hualin-adsk: Thank you! I got it.

i18n/enu/out/debug.i18n.json
```diff
@@ -1,10 +1,10 @@
 {
-	"autolispext.debug.nodap": "doesn’t exist. Verify that the file exists in the same folder as that for the product specified in the launch.json file.", 
-	"autolispext.debug.noacad": "doesn’t exist. Verify and correct the folder path to the product executable.", 
-	"autolispext.debug.acad.nosupport": "This instance of AutoCAD doesn’t support debugging AutoLISP files, use a release later than AutoCAD 2020.",
-	"autolispext.debug.launchjson.path": "Specify the absolute path to the product with the Path attribute of the launch.json file.",
-	"autolispext.debug.prod.path.win":"Specify the absolute path for the product. For example, C://Program Files//Autodesk//AutoCAD//acad.exe.",
-	"autolispext.debug.prod.path.osx": "Specify the absolute path for the product. For example, /Applications/Autodesk/AutoCAD.app/Contents/MacOS/AutoCAD.",
-	"autolispext.debug.prod.path.other":"Specify the absolute path for the product.",
-	"autolispext.debug.noprocess.eror":"No process for which to attach could be found."
+  "autolispext.debug.nodap": "doesn’t exist. Verify that the file exists in the same folder as that for the product specified in the launch.json file.",
```
hualin-adsk, line 2: it looks like it removes the json format, so it has so much file changes.

JD-Howard: I would say it fixes the JSON format because it is all ran through a single standardized JSON formatter; Newtonsoft. The data was essentially read and then written back with just as it was (value wise) but with some structural normalization. 

You'll notice some files had this normalization performed and other didn't. This is because our files had inconsistencies. Now they don't.

utils/autolispext.i18.helper/MainWindow.xaml.cs
```diff
@@ -0,0 +1,293 @@
...
+                            jsonFiles.Add(expected, new Dictionary<string, string>());
+
+                        string cv = jsonFiles[expected].ContainsKey(localContent) ? jsonFiles[expected][localContent] : "";
+
+                        if (References.ContainsKey(localContent) == false)
+                            References.Add(localContent, new List<localizeReference>());
+
+                        if (References[localContent].Count != Shared.codes.Count)
+                            References[localContent].Add(new localizeReference(new WeakReference<tsFileReference>(this), langid, expected, localContent, cv));
+                    }
```
JD-Howard, line 276: This is the section where the JSON files are either read using the UnPack extension method (if they exist and more specifically 265) into a simple `Dictionary<string, string>` or an empty `Dictionary<string, string>` is created (line 267) if they don't currently exist. IE, a new *.TS file was made with a brand new `AutoLispExt.localize()` reference. These values also have UI element representations (localizedReference objects) that will eventually overwrite the JSON Dictionary values using key value injection if the user clicks (line 106) save. I did it in this "cluster" of a way way so that only the found TypeScript references were ever injected and no orphaned or undetected localizations could ever be erased.

i18n/enu/out/debug.i18n.json
hualin-adsk: ok, and it (newtonsoft) get the same output as this extensions? https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

JD-Howard: I found [this issue](https://github.com/prettier/prettier/issues/8815) talking about how Prettier defaults to using UTF-16 and that is what I believe Newtonsoft defaults to using. So, encoding should be a non-issue. From the structural aspect, I am fairly certain that is extremely configurable (even the encoding) within Prettier and thus highly subjective. So... I guess that's a yes...

In general the JSON files in question (whole document) are exceptionally primitive (`Dictionary<string, string>`) and no "release" worthy version of a JSON reader should ever misinterpret these. The only real risk should be the encoding or human failures in maintaining the structures that corrupt the read process. I did not test for the human error side, but this app would actually help prevent that. I don't think the encoding could have been screwed up or else I would think Git would have flagged the entirety of every document instead of the structural spacing parts.

i18n/fra/out/project/openProject.i18n.json
```diff
@@ -2,5 +2,6 @@
   "autolispext.project.openproject.onlyprjallowed": "Seuls les fichiers PRJ sont autorisés.",
   "autolispext.project.openproject.readfailed": "Impossible de lire le fichier de projet: ",
   "autolispext.project.openproject.malformedfile": "Fichier de projet incorrect: ",
-  "autolispext.project.openproject.label": "Ouvrir un projet"
+  "autolispext.project.openproject.label": "Ouvrir un projet",
+  "autolispext.project.openproject.nospaces": ""
```
JD-Howard, line 6: As much as you see a lot whole documents having spacing adjustments, you will also find a lot that don't and that is why I say this "fixed" the formatting my normalizing it.

i18n/enu/out/debug.i18n.json
hualin-adsk: Thanks! I only don't want to add/remove the blank spaces in different PRs because of the different json formatters.

hualin-adsk (approved):

PR #45, opened 2020-11-08
F12 GoToDefinition Provider

JD-Howard:
**Objective**
Add the ability to use the 'native' VSCode _Go To Definition / F12_ features from our extension. This has a preference to use open files and PRJ references whenever possible, but it will fallback to use the entirety of the workspace documents if not found in the preferred references.

- Conceptually the document management portion of the extension has now evolved to pre-load the workspace ReadonlyDocuments and stop setting them to null. On 10MB of LSP's, it is only a few seconds to read all the data and create the basic representation. In addition to this overhead, I am also pre-loading the atomsForest. Which has significantly more overhead than the base document represenentations, but this was done in every way I know how to mitigate the impacts. As it stands, until the workspace is fully loaded, the PRJ palette is in a 'busy' state. My stress testing scenarios represented extremely unrealistic circumstances and should be very reasonable to the vast majority of users. All of this was done because it would now requires enough load time (at the first F12) to make users question if VSCode is crashing.
- No language localization's were required for F12 because VSCode natively reports a tool tip when no results are found. There is one added, but unrelated "workspace finished loading" information message I will go localize just before it looks like this is approved; keeping the noise down for now.
- This was hooked up for Defun & Defun-Q functions. I am on the fence if it makes any sense at all to extend this to locate the declaration or Setq locations of a variable name. I would think the only utility for that would be actual declaration headers and global (top level) setq variable setters.
- fixed a bug in the ReadonlyDocument.getMemoryDocument() method. Was missing the fileName property setter.
- Expanded the functionality of the Sexpression class with new functions and relocated an Sexpression specific method (DOM Selector like function) out of the ReadonlyDocument into Sexpression because it was more likely to get used as a dot function than just knowing it exists in the ReadonlyDocument somewhere.
- The `executeCommand('editor.action.addSelectionToNextFindMatch')` trick to find the below symbol would in fact find the symbol, but somehow blocked the UI from immediately performing the GoTo  command. I had to figure out a different way to get this information (without regex) that didn't upset the UI.
- Created providers folder for these small 1-off's. The F2 renaming provider is expected to go in there next. I can move it out of there if you want, but let me know where please.

**Tests Passed**
- Tests were performed on 10mb of *.LSP's in workspace. Memory implications were documented in code comments.
- The complete workspace load (with atomsForest) on 10mb of *.LSP files was about 20-30 seconds, but the **only** noticeable blocking that occurred while performing this was a "busy" PRJ palette.
- verified Opened & PRJ references get preference
- wrote test suite cases for a number of the integral functions backing this feature up.
- Added a more complex sample LSP and threw a bunch of collection scenarios at it. Infinitely more than I have permanent tests for. There is an entire class not part of this PR designed to blueprint every defun, variable and function reference in an LSP. I am generally happy with it, but decided to extract just things that made it work and expand existing classes instead.

**Screenshots**
Didn't bother because we all know what F12 does


extension/src/providers/gotoProvider.ts
```diff
@@ -0,0 +1,67 @@
...
+	} 
+
+	// This was designed around the idea of "preference" for opened and workspace documents
+	private findGoToLocations(searchFor: string): vscode.Location[] {
+		const locations: vscode.Location[] = [];
+		let possible = this.findDefunMatches(searchFor, AutoLispExt.Documents.OpenedDocuments.concat(AutoLispExt.Documents.ProjectDocuments));
+		if (possible.length === 0) {
+			possible = this.findDefunMatches(searchFor, AutoLispExt.Documents.WorkspaceDocuments);
+		}
+		possible.forEach(item => {
```
hualin-adsk, line 40: If the opened files are not in current workspace, and are not included in the project file, it will not find in those files?

extension/src/providers/gotoProvider.ts
```diff
@@ -0,0 +1,67 @@
...
+		});	
+		return locations;
+	}
+
+	private findDefunMatches(searchFor: string, searchIn: ReadonlyDocument[]): LocatedItem[] {
+		const result: LocatedItem[] = [];
+		searchIn.forEach((doc: ReadonlyDocument) => {
+			doc.atomsForest.forEach(atom => {
+				if (atom instanceof Sexpression){
+					const defs = atom.findChildren(/^DEFUN$|^DEFUN-Q$/i, true);
```
hualin-adsk, line 51: It looks like we only take care the function. Maybe to take care the global (`setq`) or function local (setq) is also needed.

```
(defun sampleFunc (x y / a b c d)
  (setq a (list 1 2 3 4) ; does this screw it all up?
        d (setq b 0)
        b (mapcar '+ a))
  (foreach x a
    (setq d (1+ d)) ;; select d and hit F12, maybe it should goto the first line
  )
```

extension/src/providers/gotoProvider.ts
JD-Howard: I am doing a concat on the `AutoLispExt.Documents.OpenedDocuments` and `AutoLispExt.Documents.ProjectDocuments` lists, so I believe these will find non-workspace documents as long as they have been "activated" in the interface. If the tab was left open from a previous sessions, was not in the workspace and has not been activated in the new session, then the `AutoLispExt.Documents.OpenedDocuments` variable would not know about the document. Agreed, there are scenarios where it would not be useable, but I don't think we can do anything about them until the VSCode team provide access to what is actually opened in the UI.

extension/src/providers/gotoProvider.ts
JD-Howard: I was thinking that too, but then I got concerned the implications of doing that could easily block simpler function names from ever being located in the workspace. I guess if I check to see their Sexpression.Atoms[1].symbol is the match it could prefer functions and if that is false it can prefer variables headers/declarations.

extension/src/providers/gotoProvider.ts
```diff
@@ -0,0 +1,67 @@
...
+		try {
+			return this.findGoToLocations(selected);
+		} catch (error) {
+			return;	// I don't believe this requires a localized error since VSCode has a default "no definition found" response
+		}
+	} 
+
+	// This was designed around the idea of "preference" for opened and workspace documents
+	private findGoToLocations(searchFor: string): vscode.Location[] {
+		const locations: vscode.Location[] = [];
```
hualin-adsk, line 35: If the `searchFor` is built-in symbols like `mapcar`, it should return it early?

JD-Howard: Correct me if I am wrong, but all (or many) of those symbols could theoretically be redefined. If that is true, then the internally defined symbols don't necessarily mean anything from an _exclude_ perspective. If that is not true, then we should probably only exclude the `AutoLispExt.Data.internalLispFuncs` symbols.

hualin-adsk: I am not sure this point. If the performance is not an issue, it is ok to do F12 on built-in symbols.

JD-Howard: Here is what I am talking about. AutoCAD doesn't explicitly protect those built in symbols, they could all theoretically have alternate definitions.
image

extension/src/documents.ts
```diff
@@ -103,14 +103,26 @@ export class DocumentManager{
...
+					this._workspace.set(fileUri.fsPath, ReadonlyDocument.open(fileUri.fsPath));
 				}
 			});
+		}).then(() => {
+			AutoLispExt.Documents.WorkspaceDocuments.forEach(d => {
+				// pre-caching the AtomsForest is a more significant memory jump because of its parent/child structure
+				// However, on 10mb's of LSP's in the workspace, this is only about a 90mb sustained jump and adds a lot of performance
+				// in other areas that makes this worth the cost to passively (non-blocking) pre-load the data.
+				// Also note, generating the pre-atomsForest does cause a massive memory spike (500mb) until the workspace is fully loaded.
+				d.updateAtomsForest();
```
hualin-adsk, line 121: In my local env the F12 can not work to switch to your branch, and I see it is busy in this call stack:
```
<anonymous> (c:\Program Files\Microsoft VS Code\resources\app\out\bootstrap-fork.js:5)
o (c:\Program Files\Microsoft VS Code\resources\app\out\bootstrap-fork.js:5)
<anonymous> (c:\Program Files\Microsoft VS Code\resources\app\out\bootstrap-fork.js:5)
tokenizeString (e:\test\AutoLispExt\extension\src\format\parser.ts:131)
<anonymous> (e:\test\AutoLispExt\extension\src\project\readOnlyDocument.ts:131)
<anonymous> (e:\test\AutoLispExt\out\project\readOnlyDocument.js:8)
<anonymous> (e:\test\AutoLispExt\out\project\readOnlyDocument.js:4)
updateAtomsForest (e:\test\AutoLispExt\out\project\readOnlyDocument.js:117)
<anonymous> (e:\test\AutoLispExt\extension\src\documents.ts:121)
<anonymous> (e:\test\AutoLispExt\extension\src\documents.ts:116)
processTicksAndRejections (Unknown Source:94)
Promise.then (Unknown Source:1)
initialize (e:\test\AutoLispExt\extension\src\documents.ts:115)
DocumentManager (e:\test\AutoLispExt\extension\src\documents.ts:259)
initialize (e:\test\AutoLispExt\extension\src\context.ts:39)
activate (e:\test\AutoLispExt\extension\src\extension.ts:25)
_callActivateOptional (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:933)
_callActivate (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:933)
<anonymous> (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:932)
processTicksAndRejections (Unknown Source:94)
```

JD-Howard: Could you help me understand what you mean by busy? I am actively trying to implement the variable definition side of this. Could that be the problem?

JD-Howard:
I just added the ability to F12 a variable. Which included some additional Sexpression enhancements (and tests) for stepping upward through an Sexpression tree.  The concept was:
- first check to see if it is used like a function, if it was then don't treat it like a variable and only look for function definitions
  - may need to add something that checks to see if its contained in a variable header, but I'll have to take care of that later tonight.
- else find closest declaration. This could have been a Foreach, Vlax-For, Lambda or Defun. If there are nested scenarios or (for example) a foreach used a variable name and it was also localized by the defun, it would first jump to the foreach and from the foreach it would jump up to the defun
  - If none of that pans out it throws a hail marry and gets the first 1st teir Setq statements from the ReadonlyDocuments atomsForest nodes and will attempt to locate global/shared defintions of that variable.

extension/src/documents.ts
JD-Howard: I am going to guess there may be a bad document in your workspace. It looks to me like the error is being generated while trying to generate the atomsForest on the ReadonlyDocument. Since I don't have this problem, could you please surround line 121 of `documents.ts` with a Try/Catch and tell me what scenario is causing this; email me the document maybe. I'll gladly fix it if I had something to go on, but looking at the updateAtomsForest function, it already has the stuff I was worried about in a Try/Catch.

hualin-adsk: See this pics(and let me know if you need the LSP file I used):
F12issue

hualin-adsk: And check your email for the lisp file

hualin-adsk: Thanks Josh!
It looks like I hit the performence issue here. I opened a folder which contains about 30 lisp files. I will test to open a .prj file which referece 40 lisp files, and that is a real product source codes.

extension/src/providers/gotoProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+import { ReadonlyDocument } from '../project/readOnlyDocument';
+
+interface LocatedItem {
+	line: number;
+	column: number;
+	fspath: string;
+}
+
+export class AutolispDefinitionProvider implements vscode.DefinitionProvider{
+	async provideDefinition(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken): Promise<vscode.Location | vscode.Location[]> {
```
hualin-adsk, line 13: This is very cool feature!!!
I test a little and it works cool for functions and variables!

extension/src/documents.ts
JD-Howard: I was quite concerned about the performance issues, but I thought these were passively loading in the background and shouldn't cause anything other than a blocked Project Explorer. Can you share your computer specs? I have 480 LSP files around 10mb's and the entire workspace finishes loading in (at most) 30 seconds for me. The first part that just makes the ReadonlyDocuments (you can see something in the status bar) is less than 10 seconds. We can go back to the demand-load at the first F12, but that will lock the user interface long enough that people may think they are crashing.

Note: PRJ references are demand loaded, so you could be in for a wait on those if it has many large references.

JD-Howard: I tested this by deleting an open parenthesis. It does break a degree of its abilities, but I think that is working as intended because it can't properly analyze/navigate the code. No noticeable crashes though.

FYI, my specs: i7-8086k, 32GB ram, 2TB SSD primary. However, the workspace folder I use is on a NAS that uses traditional spin drives for long term storage and a 0.5TB SSD's for memory resident (recently used) files. So, I am at least slowed a bit by networking.

hualin-adsk: this is my machine configuration:
image

JD-Howard: My processor runs at 4ghz, but I would think your dual processors would match it. So, I am not quite sure why 30 files would take any time at all. Theoretically, if I can do 480 in 30s, then I would think yours would be handling 30 in 5s or less. Its concerning because I've based the performance on a large scale thinking I was covering the bases and maybe that isn't the case. Is your workspace particularly large with non LSP's by chance? That is about the only thing I can think of that could be causing this massive disparity. Mine has a bit of other stuff, but the bulk is definitively LSP files.

JD-Howard: Here is an idea. Would if, we go back to demand loading the atomsForest, but do the F12 workspace (specifically) searches with a brute force `/(DEFUN|DEFUN-Q)TheirSelection/ig.test(document.fileContent.replace(/\s/g))` to detect if its even worth generating the atomsForest in the first place. That should theoretically stop the unnecessary generation of every documents atomsForest except when they legit come into scope. This should have performance implications for the PRJ's as well. Thoughts?

JD-Howard:
The last commit (eeb8a7a) added another Sexpression navigation helper function and a test for that function. Removed the largely unnecessary interface, I think it was put there because of a different plan that didn't unfold. Also handled a number of edge cases. I think what I have is fine for symbol definitions, but the atomsForest returning `String|Sexpression` is going to pose real complications for the F2 renaming I may head for next now that the tools are in place to better understand scope. Other than a single reference in _formatter.ts_ `formatGut()` I am the primary user of this. Anyone know of a reason why the `formatGut()` function couldn't use a `LispAtom|Sexpression` instead?

extension/src/format/sexpression.ts
```diff
@@ -890,4 +890,94 @@ export class Sexpression extends LispAtom {
...
+                return index;               
+            } else {
+                return this.nextKeyIndex(index, forSetq);
+            }
+        } else {
+            return -1;
+        }
+    }
+
+    getNthKeyAtom(count: number){
```
JD-Howard, line 950: I am actually not sure what to call this variable. It's not an index and its not really a count either. Its closer to an index, but an index specific to the LispAtoms/Sexpressions that have actual significance.

extension/src/documents.ts
hualin-adsk: I agree that it should cache there, maybe it needs to find a async way to do cache.
I test it in a real product, it look like the performence is about 5~10 seconds, and it doesn't find the one I expected:
bigprj

JD-Howard: PRJ's are demand loaded and that hasn't been tested on the large scale your talking about. Later tonight I will create some test scenarios and see if I can replicate the problem. Probably test my pre-check idea's viability too. At the very least it should be helpful for identifying things that should have found something that ultimately didn't. **Are the LSP's inside the PRJ also in your workspace or do they just exist in the PRJ?** I'll test both, I am just curious about your situation.

I did add the async decoration to the function that initializes the atomsForest, I suppose that not being called with an await could be causing it to gloss over the data pull. Maybe that needs to be removed.

hualin-adsk: For the second GIF I only open a prj file, and there are no workspace opened and no folder opened.
And for the function `empty_trmlst` it has function body, but F12 can not find it. This is also weird.

I am also narrowdown the issue here. And I think for a product prj file it is normal to have hundreds of source files, as the issue https://github.com/Autodesk-AutoCAD/AutoLispExt/issues/47 mentioned.

extension/src/project/projectTree.ts
```diff
@@ -123,7 +123,7 @@ export class ProjectTreeProvider implements vscode.TreeDataProvider<DisplayNode>
     public updateData(newRootNode: ProjectNode) {
         this.rootNode = newRootNode;
         vscode.commands.executeCommand("setContext", "autolisp.hasProject", ProjectTreeProvider.hasProjectOpened());
-        this.onChanged.fire(this.rootNode);
+        this.refreshData(null);
```
JD-Howard, line 126: not sure why it picked this up I went and copied the line from the bug fix #46

extension/src/project/projectTree.ts
```diff
@@ -236,5 +236,6 @@ export function addLispFileNode2ProjectTree(root: ProjectNode, fileName: string,
     fileNode.fileExists = fs.existsSync(fileName);
     fileNode.rawFilePath = rawFilePath;
     fileNode.document = ReadonlyDocument.open(fileName);
-    root.sourceFiles.push(fileNode);
+    fileNode.document.updateAtomsForest();
+    root.sourceFiles.push(fileNode);
```
JD-Howard, line 240: Added a pre-load on the PRJ atoms forest since it is an explicit act to load the project and makes sense for them to suffer a load time vs the entire workspace. From what you were saying, I can expect these to potentially be very large.

extension/src/documents.ts
```diff
@@ -103,15 +103,29 @@ export class DocumentManager{
...
-		// 		This feature does make the "fair assumption" that AutoCAD machines has plenty of memory to be holding all this information
-		// 		This was stress tested with a root workspace folder containing 10mb of *.LSP files and the memory footprint increased less than 50mb
+		// 		This feature does make the "fair assumption" that AutoCAD machines have plenty of memory to be holding all this information
+		// 		The impact of creating read-only documents was stress tested with a root workspace folder containing 10mb of *.LSP files
+		//		and the memory footprint from the readonlyDocument's increased the memory (sustained) by less than 50mb		
 		vscode.workspace.findFiles("**").then((items: vscode.Uri[]) => {
 			items.forEach((fileUri: vscode.Uri) => {			
 				if (this.getSelectorType(fileUri.fsPath) === DocumentManager.Selectors.lsp) {
-					this._workspace.set(fileUri.fsPath, null); // initialized as null to blueprint the workspace files without the overhead of reading their contents
+					this._workspace.set(fileUri.fsPath, ReadonlyDocument.open(fileUri.fsPath));
```
JD-Howard, line 112: This is the only pre-load happening as of pre-check method (17a9977) and for me this is 4-10 seconds on a 10mb of LSP's.

extension/src/documents.ts
JD-Howard: I have now tested this with a non-workspace and PRJ. It worked for me. I also tested this with a workspace and a PRJ containing the same files, but from another folder outside of the workspace. It successfully opened the PRJ version as the result. In short, give it another try and lets see if we have a winner now.

hualin-adsk: To get the latest codes I tried to open the large prj files, it get hang there:
```
get line (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:513)
offsetAt (e:\test\AutoLispExt\extension\src\project\readOnlyDocument.ts:185)
validateRange (e:\test\AutoLispExt\extension\src\project\readOnlyDocument.ts:255)
getText (e:\test\AutoLispExt\extension\src\project\readOnlyDocument.ts:239)
readComments (e:\test\AutoLispExt\extension\src\format\parser.ts:62)
tokenizeString (e:\test\AutoLispExt\extension\src\format\parser.ts:99)
updateAtomsForest (e:\test\AutoLispExt\extension\src\project\readOnlyDocument.ts:131)
addLispFileNode2ProjectTree (e:\test\AutoLispExt\extension\src\project\projectTree.ts:239)
ParseProjectDocument (e:\test\AutoLispExt\extension\src\project\openProject.ts:126)
OpenProjectFile (e:\test\AutoLispExt\extension\src\project\openProject.ts:43)
<anonymous> (e:\test\AutoLispExt\extension\src\project\openProject.ts:28)
fulfilled (e:\test\AutoLispExt\out\project\openProject.js:5)
processTicksAndRejections (Unknown Source:94)
Promise.then (Unknown Source:1)
step (e:\test\AutoLispExt\out\project\openProject.js:7)
fulfilled (e:\test\AutoLispExt\out\project\openProject.js:5)
processTicksAndRejections (Unknown Source:94)
Promise.then (Unknown Source:1)
step (e:\test\AutoLispExt\out\project\openProject.js:7)
<anonymous> (e:\test\AutoLispExt\out\project\openProject.js:8)
```

JD-Howard: So, that error is happening while loading the atomsForest, but within code I didn't write. I can not pre-load the atomsforest to protect the integrity of the load, but I assume that won't ultimately solve the problem. I don't expect you to email me your full program, but there is clearly something in there the parser doesn't like. Please try to figure out what that is.

JD-Howard: I am going to guess that Line:62 of parser.ts is putting a null value into `readComments()` and that is bubbling up to the `offsetAt()`
I'd suggest adding a conditional breakpoint on line 62 !(startPos2d && endPos2d)

hualin-adsk: I am narrowing down the issue. maybe I can take some time this weekend.
it looks like for each F12 it build the S-expression each time? it looks like `getMemoryDocument` return a new memory document.

JD-Howard: getMemoryDocument kind of 'casts' the active TextDocument to a ReadonlyDocument so it is in the same context as everything else. It would do this same operation for anything in the `AutoLispExt.Documents.OpenedDocuments` getter. It kind of has to if we want to insure that unsaved edits are captured. For the same reasons, the atomsForest must be regenerated. 
Anything that is not in an opened state would already have a ReadonlyDocument representation. This would include the PRJ & workspace files. The workspace files are no longer pre-loading the atomsForest, but the PRJ's are. I did this because a load operation when opening a PRJ seems like a much more explicit intent, a load time on open makes some degree of sense from a user perspective and you told me we could expect these to potentially be excessively large; which only substantiates allowing a load time from my perspective. We can shut this off, but this problem will not go away. There is very clearly a bug in the parser we need to track down.

hualin-adsk: I test the following case:

1.  start clean vscode
2.  open 1.lsp
3.  open 2.lsp ;; in 2.lsp call a function in 1.lsp
4.  F12 on the function call in 2.lsp
5.  do the above step 2 times

I think after the step5 it should no longer call the parser to parse the 1.lsp file. And the 1.lsp atoms forest are cached.
But the codes is not that expected: https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/f5d8e754065dafef1560cd034a2c77cb370fae49/extension/src/project/readOnlyDocument.ts#L75 create the ReadOnlyDocument each time, so there is no cache for atoms forest:
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/f5d8e754065dafef1560cd034a2c77cb370fae49/extension/src/project/readOnlyDocument.ts#L109

So it calls the https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/f5d8e754065dafef1560cd034a2c77cb370fae49/extension/src/project/readOnlyDocument.ts#L112 every time.

And I tried to create a project file to contains above 1.lsp and 2.lsp, and only open 2.lsp file, it looks like this code:
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/f5d8e754065dafef1560cd034a2c77cb370fae49/extension/src/documents.ts#L57
The un-opened 1.lsp is also cached by Vscode document instead of the ReadOnlyDocument:
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/f5d8e754065dafef1560cd034a2c77cb370fae49/extension/src/documents.ts#L9

hualin-adsk:
Hi Josh,

Thank you for your patience on this long reviewing! I want to take more time on this to address the performance issue, because the DocumentManager is the infrastructure codes, some advanced feature will depends on it, for example F12 you did, find all references, Rename, and so on.

thanks

extension/src/documents.ts
JD-Howard: - Your first scenario is dealing with documents that are Opened. We have no idea if there are unsaved edits and thus the only safe thing to do is regenerate from the TextDocument.
  - In the context of a rename operation it could be extremely destructive if we didn't regenerate.
  - If the document is opened I don't believe caching is a safe or feature complete option. Unsaved means we don't know if they've been working on it for the last 2 hours and just haven't saved.
- Your second scenario is because PRJ's have the opposite slash direction evenything else provides. Easy enough to normalize all of these; just noticed this the other day. Because they aren't 1-1 it never finds the existing ReadonlyDocument and keeps regenerating.

JD-Howard:
One of the best things you could do to increase performance in opened documents is split 8000 line LSP's into several documents. I went and looked at my largest files and they are all less than 2500 lines and I'd honestly say anything that gets up around 1000 starts to become unmanageable. I say that because the current arrangement does a very speedy pre-check that should help it not evaluate a good chunk of them. You put 8000 lines in a single (opened) document and its always going to be relevant and always need to be regenerated.

hualin-adsk:
I know that to have 8000 lines codes in one file is rare, but it is a **real world** cases in my hand.

For some small project it works very well, for example I tried this open source poject: https://github.com/jdsandifer/AutoLISp

extension/src/documents.ts
hualin-adsk: So for the second scenario it looks like a bug?
And I just realized that the https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/main/extension/src/documents.ts#L142 is only for `workspace` files.

Maybe in the class `DocumentManager` it should always cache the `ReadonlyDocument` for **all** files (workspace, and prj files, even the current file). When cache the ReadonlyDocument, it also caches the changed time stamp of this file; when to get the cache, try to check the changed time stamp, if it is out of date, it regennerates the atoms forest.  Do you think this is more cheaper way?

extension/src/documents.ts
hualin-adsk: I also didn't get slow feeling to open a workspace which contains lots of files

extension/src/format/sexpression.ts
```diff
@@ -890,4 +890,94 @@ export class Sexpression extends LispAtom {
...
+    }
+
+
+    // This version was necessary to properly alternate over SETQ Name vs Value 
+    private isValidForSetq(atom: LispAtom) {
+        return !atom.isComment() && !['\'', '(', ')', '.'].includes(atom.symbol) && (atom instanceof Sexpression || atom.symbol.trim().length > 0);
+    }    
+    // This is general purpose utility to make sure primitives such as strings, numbers and decorations are not evaluated
+    private isValidNonPrimitive(atom: LispAtom) {        
+        return !atom.isComment() && !['\'', '(', ')', '.'].includes(atom.symbol) && !(/^".*"$/.test(atom.symbol)) && !(/^\'{0,1}\-{0,1}\d+[eE\-]{0,2}\d*$/.test(atom.symbol));
```
hualin-adsk, line 930: maybe give more meanful name for the complex `test`
https://www.avdweb.nl/arduino/programming/descriptive-variable-names

extension/src/format/sexpression.ts
```diff
@@ -890,4 +890,94 @@ export class Sexpression extends LispAtom {
...
+    }
+
+
+    // This is an ittoration helper that makes sure only useful LispAtom/Sexpression's get used for contextual data collection purposes.
+    nextKeyIndex(currentIndex: number, forSetq?: boolean): number {
+        let index = currentIndex + 1;
+        if (index < this.atoms.length) {
+            const atom = this.atoms[index];
```
hualin-adsk, line 938: very nice `const`!

extension/src/providers/gotoProvider.ts
```diff
@@ -0,0 +1,145 @@
...
+			rDoc.atomsForest.forEach(sexp => {
+				if (sexp instanceof Sexpression && sexp.contains(position)){
+					const atom = sexp.getAtomFromPos(position);
+					if (atom) {
+						selected = atom.symbol.replace(/^[']*/, '');
+					}
+				}
+			});
+		}
+		if (['(', ')', '\'', '.'].includes(selected)){
```
hualin-adsk, line 24: maybe add` ;`

extension/src/providers/gotoProvider.ts
```diff
@@ -0,0 +1,145 @@
...
+	
+	private findFirstVariableMatch(doc: ReadonlyDocument, start: vscode.Position, searchFor: string, searchIn: Sexpression): vscode.Location[] {
+		const result: vscode.Location[] = [];
+		const ucName = searchFor.toUpperCase();
+		let context = searchIn.getSexpressionFromPos(start);
+		let flag = true;
+		do {
+			const parent = searchIn.getParentOfSexpression(context);
+			const atom = parent?.getNthKeyAtom(0);
+			if (atom && /^DEFUN$|^DEFUN-Q|^LAMBDA$/i.test(atom.symbol)) {
```
hualin-adsk, line 80: It is interesting that this is my first see this form `/^DEFUN$|^DEFUN-Q|^LAMBDA$/i `

extension/src/providers/gotoProvider.ts
```diff
@@ -0,0 +1,145 @@
...
+					result.push(new vscode.Location(vscode.Uri.file(doc.fileName), new vscode.Position(tmpVar.line, tmpVar.column)));
+				}
+			}
+			if (!parent || result.length > 0){
+				flag = false;
+			} else {
+				context = parent;
+			}
+		} while (flag);
+		// If we still haven't found anything check the 1st occurrence of setq's. This will find globals setqs and nested ones possibly inside other defuns
```
hualin-adsk, line 106: it is best to wrap to a small function

extension/src/providers/gotoProvider.ts
```diff
@@ -0,0 +1,145 @@
...
+			});
+		}
+		return result;
+	}
+
+	private findDefunMatches(searchFor: string, searchIn: ReadonlyDocument[]): vscode.Location[] {
+		const result: vscode.Location[] = [];
+		const regx = new RegExp('\\((DEFUN|DEFUN-Q)' + escapeRegExp(searchFor) + '\\(', 'ig');
+		searchIn.forEach((doc: ReadonlyDocument) => {
+			if (regx.test(doc.fileContent.replace(/\s/g, ''))){
```
hualin-adsk, line 128: what is the \s/g

extension/src/documents.ts
JD-Howard: On the ReadonlyDocument side, we have a direct fileContent property to read. So, I think the TextDocument.getText() on a max-range is cheap enough as long as we don't have resulting EOL comparison problems. 

I'll go make some more tweaks to recycle ROD's and fix the slashes, but I am going to leave PRJ isolated to the project for now. I really don't want the Project manager to be in constant contact with the Document manager if it can be avoided. Right now they have a very substantial separation of concerns and if we don't need to muddy those waters, then we really shouldn't.

extension/src/providers/gotoProvider.ts
JD-Howard: Is that good or bad? I see I missed a $ at the end of DEFUN-Q. What do you normally do to say something has to _completely_ match? I know they would need to be in parenthesis if there are "other" pieces I was trying to also match.

extension/src/providers/gotoProvider.ts
JD-Howard: I am attempting to evaluate this without the context of tabs, line feeds and spaces. Basically if someone decided to they like this format declaration it won't break the primitive search:

```lisp
(defun
       whatever 
       (a b / c d e)
  (princ (setq c (+ a b)))
)
```

extension/src/providers/gotoProvider.ts
JD-Howard: I had started playing with the renameProvider and did begin breaking some stuff out that could be recycled. Never made it into this chunk of code, but I will look into it.

extension/src/providers/gotoProvider.ts
JD-Howard: Thank you, I will certainly add this to avoid unnecessary work.

On some level, I wish AutoLisp had more restrictive naming so we could just use \W to exclude all singular symbols, but sadly `(defun # () (princ "yo!"))` is a perfectly valid function definition. I actually use that one for a custom string interpolation function actually.

extension/src/format/sexpression.ts
JD-Howard: How about `notNumberStringOrProtected()` ?

test_case/pdfMarkups.lsp
```diff
@@ -0,0 +1,162 @@
+; random sample file
+
+(defun sampleFunc (x y / a b c d)
```
hualin-adsk, line 3: If the test case codes is from other open source project, please check its license.

test_case/pdfMarkups.lsp
```diff
@@ -0,0 +1,162 @@
+; random sample file
```
hualin-adsk, line 1: and I think this `test_case` should exclude in .vscodeignore file

extension/src/providers/gotoProvider.ts
hualin-adsk: It is good! I think maybe give more readable name for this regex `/^DEFUN$|^DEFUN-Q|^LAMBDA$/i`

test_case/pdfMarkups.lsp
JD-Howard: This isn't an issue, I pulled something out of my code base and it is completely non-functional without highly unique DLL's that I've made to perform PDF I/O and SharePoint operations from within AutoCAD.

extension/src/documents.ts
hualin-adsk: `but I am going to leave PRJ isolated to the project for now.`
I test that the above scenario is not an issue for slash. The two files are opened, so it always regenerate the atoms forest. I think after refining the ReadonlyDocument map, for each lisp file it only needs to regenerate atoms one time if it is not touched. (Also notice that if the opened files are also in project files, it will regenerate atoms twice: https://github.com/JD-Howard/AutoLispExt/blob/roDocSymbolMap/extension/src/providers/gotoProvider.ts#L53)

extension/src/providers/gotoProvider.ts
hualin-adsk: Maybe add to check if the current atom is in comment, avoid to do work.

JD-Howard:
Okay, I think I did everything in this that was discussed, except what is expected to be handle in a different PR that updates the DocumentManager. 

_providerShared.ts_ is a new file I made it to offload stuff that will be used by multiple (F2, F12 & References) providers into a common place; IE, just too confusing to have pointers to static methods within completely different providers. So, they were moved out to a common file and I suspect more will get populated as the other two providers are built out.

extension/src/providers/providerShared.ts
```diff
@@ -0,0 +1,40 @@
+import { Sexpression } from '../format/sexpression';
+import { ReadonlyDocument } from '../project/readOnlyDocument';
+import * as vscode from	'vscode';
+
+export namespace SearchPatterns {
+	export const LOCALIZES = /^DEFUN$|^DEFUN-Q$|^LAMBDA$/i;
```
hualin-adsk, line 6: it should be `CALLABLE`?

extension/src/providers/gotoProvider.ts
```diff
@@ -4,60 +4,40 @@ import { AutoLispExt } from '../extension';
...
-
+import { SearchPatterns, SearchHandlers } from './providerShared';
 
 export class AutolispDefinitionProvider implements vscode.DefinitionProvider{
 	async provideDefinition(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken): Promise<vscode.Location | vscode.Location[]> {
 		const editor: vscode.TextEditor = vscode.window.activeTextEditor;
 		const rDoc = ReadonlyDocument.getMemoryDocument(document);
-		let selected: string =  editor.document.getText(editor.selection);
+		let selected: string = editor.document.getText(editor.selection);
 		if (selected === '') {
```
hualin-adsk, line 14: I think it needs to remove this checking, it should always to check if it is in comment atom

extension/src/providers/providerShared.ts
```diff
@@ -0,0 +1,40 @@
...
+	export const ASSIGNS = /^SETQ$/i;
+	export const DEFINES = /^DEFUN$|^DEFUN-Q$/i;
+	export const ALL = /^DEFUN$|^DEFUN-Q$|^LAMBDA$|^FOREACH$|^VLAX-FOR$|^SETQ$/i;
+}
+
+
+export namespace SearchHandlers {
+
+
+	export function getSelectionScopeOfWork(doc: ReadonlyDocument, position: vscode.Position, selected: string): {isFunction: boolean, parentContainer: Sexpression} {
```
hualin-adsk, line 17: Thanks for this refractoring!

hualin-adsk (approved):

extension/src/providers/gotoProvider.ts
hualin-adsk: I think there is one bug:
```
(defun testcomment () 
  (progn 
    (progn 
      (setq dwgname (nth ixx (nth 2 GBL_wd_prj_ixlst)))
    )
  )
  (progn 
    ;                        (setq dwgname "" iec_p "") ;;F12 on dwgname 
  )
)
```

JD-Howard: I agree it shouldn't be working with comments, but we are talking about situation that someone visually saw it was a comment, chose to completely highlight it and to hit F12. Does that not suggest it did what they were hoping it would do for them?

extension/src/providers/providerShared.ts
JD-Howard: Thanks for the perspective, I definitively struggled with that one. I settled on LOCALIZES because the one thing they had in common was a formal variable localization header. Maybe this one should turn into DEFINES and the current DEFINES should become CALLABLE? I only say that because LAMBDA doesn't _equally_ fit into the notion of CALLABLE without a SETQ that saves it to a variable. IE, it _could_ be defined in such a way that its recyclable/callable name, but unlikely and we aren't doing enough contextual evaluation to figure that out if it was.

extension/src/providers/gotoProvider.ts
hualin-adsk: I tried C++ codes in vscode and MS C++ extension. It ignores the comment case.

extension/src/providers/gotoProvider.ts
```diff
@@ -0,0 +1,131 @@
...
+		if (['(', ')', '\'', '.', ';'].includes(selected)){
+			return;
+		}
+		try {
+			// determine scope
+			const {isFunction, parentContainer} = SearchHandlers.getSelectionScopeOfWork(rDoc, position, selected);
+			const locations: vscode.Location[] = [];
+			if (isFunction) {
+				// This has a "preference" for opened and project documents for actual definitions, but will only handle variables on the opened document.
+				let possible = this.findDefunMatches(selected, AutoLispExt.Documents.OpenedDocuments.concat(AutoLispExt.Documents.ProjectDocuments));
```
hualin-adsk, line 33: Maybe another refinement is to do searching in the `current ` file firstly, because most of the definition is in the same file; if not found, it can search `opened, workspace, prj` files

JD-Howard: - Removed all selection checking and leaned on discovery of position atom so that comment checks would happen automatically.
- I had it check the active document first and foremost before moving onto the other tiers of contextual importance. I am not sure if you intended PRJ to be last in your list or not, but I generally kept opened/PRJ's lumped together because we don't auto-reload Projects into the interface, it is a very clear intent of contextual preference for a PRJ to present.

hualin-adsk:
Thank you very much for implementing a so cool feature!

extension/src/providers/gotoProvider.ts
```diff
@@ -30,7 +27,10 @@ export class AutolispDefinitionProvider implements vscode.DefinitionProvider{
 			const locations: vscode.Location[] = [];
 			if (isFunction) {
 				// This has a "preference" for opened and project documents for actual definitions, but will only handle variables on the opened document.
-				let possible = this.findDefunMatches(selected, AutoLispExt.Documents.OpenedDocuments.concat(AutoLispExt.Documents.ProjectDocuments));
+				let possible = this.findDefunMatches(selected, [AutoLispExt.Documents.ActiveDocument]);
+				if (possible.length === 0) {
+					possible = this.findDefunMatches(selected, AutoLispExt.Documents.OpenedDocuments.concat(AutoLispExt.Documents.ProjectDocuments));
```
hualin-adsk, line 32: `ActiveDocument` may be included in `OpenedDocuments` or `ProjectDocuments`, and maybe  `OpenedDocuments` or `ProjectDocuments` has duplicate documents. In the latter PR it looks like we can filter those duplicate files, save the time to search in the atoms forest.

Issue #44, opened 2020-11-08
Project Palette having visual tree issues


JD-Howard:
**Describe the bug**
PRJ treeview is not updating except while adding new *.lsp files.

**To Reproduce**
1. Open ANY *.prj with valid file pointers
2. no items will display, but everything else is generally correct (does contains data)
3. hit refresh, no items show up
4. add a new *.lsp file reference and they all magically appear

**Desktop (please complete the following information):**
 - OS: Win10

**Additional context**
I am relatively certain this has to do with the PR #42 changes that turned all the `onChange.fire()` events into a problem. The only other recent thing that could have potentially caused this was PR #36 but I've re-checked all the code from that and I don't see any issues. Anybody have some ideas on why this is inconsistently working?

hualin-adsk:
@Sen-real Do you have any ideals on this?

Sen-real:
> @Sen-real Do you have any ideals on this?

Didn't notice problem from the two PRs. Sometime upgrading VS Code also causes problem.

JD-Howard:
I did indeed recently get a VSCode update. However, if I don't start our extension through debugging and just use the one that exists in the store, then the TreeView works fine. Still possibly linked to an update, but slightly less likely because the public version didn't break too.

Sen-real:
fixed with https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/46

PR #46, opened 2020-11-09
Fix bug #44

Sen-real:
The bug to fix:
https://github.com/Autodesk-AutoCAD/AutoLispExt/issues/44

When firing an event to refresh the entire project panel, it only works with a null argument.


hualin-adsk (approved):

JD-Howard:
@Sen-real, 
LOL, thank you! I was starting to think I was going crazy, but that 1-liner fixed the whole problem.

JD-Howard (approved):

Issue #30, opened 2020-10-15
Document Manager Enhancement Discussion


JD-Howard:
**Is your feature request related to a problem? Please describe.**
Many of the intelligence features I want in the AutoLispExt require foraging for information across PRJ, Workspace and even on fragments of LSP files. I've already started submitting PR's to help with that, but this is where I want to do some preliminary discussion on how it will be organized and what it will provide.

**Describe the solution you'd like**
This will most likely be a class so that getters can be used to dynamically serve the best available content. This will likely be a single pre-initialized instance called DocumentManager and that document manager will probably be the only export. The constructor of the document manager is intended to create and manage all the event hooks such as:
- vscode.workspace.onDidOpenTextDocument
- vscode.workspace.onDidCloseTextDocument
- vscode.workspace.onDidRenameFiles
- vscode.workspace.onDidChangeWorkspaceFolders (has to reset the FileSystemWatcher events)
- vscode.FileSystemWatcher.onDidDelete
- vscode.FileSystemWatcher.onDidCreate
- vscode.FileSystemWatcher.onDidChange (not real sure this one will be necessary)

**Note:** The recent PR to natively add ReadonlyDocument's to the LspFileNode make it so that the PRJ events do not need to inform the DocumentManager about anything that happened in that space.

Another thing this will do is scan the workspace and find all the LSP files creating ReadonlyDocuments for each of them that get stored in a Map using their fspath as the key; this is to support multiple files with the same name. This global repo will take 3rd precedence for the content delivery system. Second precedence will be given to the PRJ referenced files and top priority will be given to opened documents; when aware of them that is. Still have a technical debt to figure out which files were already opened in the workspace reload when VSCode opened. However, once any of those file tabs get activated the above events will let us know and they will begin taking top priority.

There is performance concerns about reading the contents of all these files when our extension is first loading. It does not lock the interface, but it does in fact cause a _scanning_ bar to show and some features aren't 100% available until it completes. It doesn't take long at all, even with 10mb of LSP files in your workspace, but a bit unsightly. Because of this, I am wondering if we should make a way to create a ReadonlyDocument without the actual (content) initialize that could demand load when the user is actually doing something that tries to use that data. Another option is to create the DM Map entries with null and let the document manager demand load them when requested.

**Additional context**
The expected common access point for this will be through the ContextManager.Documents getter.

JD-Howard:
[...]

PR #49, opened 2020-11-16
Doc manager update

JD-Howard:
**Objective**
This is an update to the DocumentManager to persist/recycle more data. It was decided during the DefinitionProvider PR #45 that something had to be done about about the excessively expensive `ReadonlyDocument.atomsForest` being regenerated more often than it needed to be.

This change involves a major emphasis on tracking pairs ReadonlyDocument's & vscode.TextDocuments and performing very cheap checks to detect when the ReadonlyDocument is out of date. It uses a set of flags to specify the various contextual applications. This will inevitably make a ReadonlyDocument qualify for up to 4 queries
- Currently Active Document
- Opened
- Project
- Workspace

However, the expensive parts should now be completely recycled and very cheap checks are in place to avoid the excessive parts of that **potential** duplication. From what I can tell, nothing should ever go through all 4 in expensive ways because any one would have passed the cheap pre-check, had the expensive work done and never moved on to the others.

The original version of this did a lot better job of purging dead data, but I believe this will at least make it inaccessible. This is notable for file renames and deletions performed directly in a OS file explorer. In theory, this is removing the flags that qualify them for various queries, but not necessarily removing the reference entirely across all possible situations. I believe the actions performed in the vscode workspace itself would do proper data purging purging though.

This also removed the PRJ embedded ReadonlyDocument property. Everything is demand queried from the DocumentManager when applicable, but persisted so it can efficiently serve it back next time. The cheap pre-check added in the DefinitionProvider and the fact that ReadonlyDocument creation is also quite cheap should adequately compensate for this. Note that PRJ's being opened/closed is probably the biggest issue with persisted dead data. Since the the ProjectTree's `Instance()` is always used to produce the "_single source of truth_" of fspath keys this should not actually produce erroneous PRJ references, but it has no mechanism for getting rid of the old ones. I'd like to maintain a maximum separation between the Document Manager and the Project Manager if we can.

Removed the attempts to persist data across workspace changes. From what I can tell, this was an exercise of futility. I tried multiple tests to verify it was working and kept eventually winding up with an empty `_cached` property. For example the current arrangement of the `initialize()` method that setups the cache isn't even clearing it anymore and all references are still disappearing after any action that changes the workspace.

Honestly the _documents.ts_ is vastly more clustered than it used to be since it is now handling various edge cases and was massively changed. So, I may be easier to perform your first review without the Adds/Deletes turned on.

**Tests Passed**
I kind of just described them above by identifying the potential known issues. I did run this with the F12 on a Workspace only, a PRJ only and a Workspace/PRJ hybrid.

**Screenshots**
Not Applicable

### DIFF
--- a/extension/src/documents.ts
+++ b/extension/src/documents.ts
@@ -3,12 +3,13 @@ import * as vscode from 'vscode';
 import { ReadonlyDocument } from "./project/readOnlyDocument";
 import { AutoLispExt } from './extension';
 import { ProjectTreeProvider  } from "./project/projectTree";
+import * as fs from	'fs-extra';
 
 
 enum Origins {
-	OPENED = 'O',
-	WSPACE = 'W',
-	PROJECT = 'P'
+	OPENED,
+	WSPACE,
+	PROJECT
 }
 
 
@@ -49,29 +50,15 @@ export class DocumentManager{
 	}
 	get ActiveDocument(): ReadonlyDocument { 
 		if (vscode.window.activeTextEditor) {
-			const key = vscode.window.activeTextEditor.document.fileName.replace(/\//g, '\\');
-			if (this._cached.has(key)){
-				const sources = this._cached.get(key);
-				if (!sources.native) {
-					sources.native = vscode.window.activeTextEditor.document;
-					sources.flags.add(Origins.OPENED);
-				}
-				if (!sources.internal || !sources.internal.equal(sources.native)){
-					sources.internal = ReadonlyDocument.getMemoryDocument(sources.native);
-				}
-				return sources.internal;
-			} else {
-				const sources = DocumentSources.create(Origins.OPENED, vscode.window.activeTextEditor.document);
-				this._cached.set(key, sources);
-				return sources.internal;
-			}
+			const key = this.documentConsumeOrValidate(vscode.window.activeTextEditor.document, Origins.OPENED);			
+			return this._cached.get(key)?.internal;
 		} else {
 			return null;
 		}		
 	}
 	
 	get ActiveTextDocument(): vscode.TextDocument { 
-		return vscode.window.activeTextEditor ? vscode.window.activeTextEditor.document :  null; 
+		return vscode.window.activeTextEditor?.document;
 	}
 	get OpenedTextDocuments(): vscode.TextDocument[] { 
 		return [...this._cached.values()]
@@ -86,24 +73,10 @@ export class DocumentManager{
 		const result: string[] = [];
 		if (ProjectTreeProvider.hasProjectOpened()){
 			ProjectTreeProvider.instance().projectNode.sourceFiles.forEach(x => {
-				result.push(x.filePath.replace(/\//g, '\\'));
+				result.push(this.normalizePath(x.filePath));
 			});
 		}
 		return result;
-	}
-
-	getDocument(nDoc: vscode.TextDocument) {
-		const key = nDoc.fileName.replace(/\//g, '\\');
-		if (this._cached.has(key)) {
-			const sources = this._cached.get(key);
-			sources.native = nDoc;
-			if (!sources.internal || !sources.internal.equal(sources.native)){
-				sources.internal = ReadonlyDocument.getMemoryDocument(sources.native);
-			}
-			return sources.internal;
-		} else {
-			this._cached.set(key, DocumentSources.create(Origins.OPENED, nDoc));
-		}
 	}
 
 	// General purpose methods for identifying the scope of work for a given document type
@@ -125,30 +98,72 @@ export class DocumentManager{
 		}
 	}
 
+	private normalizePath(path: string): string {
+		return path.replace(/\//g, '\\');
+	}
+
+	private tryUpdateInternal(sources: DocumentSources){
+		if (sources.native && (!sources.internal || !sources.internal.equal(sources.native))) {
+			sources.internal = ReadonlyDocument.getMemoryDocument(sources.native);
+		}
+	}
+
+	private pathConsumeOrValidate(path: string, flag: Origins): string {
+		const key = this.normalizePath(path);		
+		if (fs.existsSync(key) && this.getSelectorType(key) === DocumentManager.Selectors.lsp) {
+			if (this._cached.has(key) === false) {
+				const sources = DocumentSources.create(flag, ReadonlyDocument.open(key));
+				this._cached.set(key, sources);
+			} else {
+				const sources = this._cached.get(key);				
+				this.tryUpdateInternal(sources);
+				sources.flags.add(flag);
+			}
+			return key;
+		} else {
+			return '';
+		}
+	}
+
+	private documentConsumeOrValidate(doc: vscode.TextDocument, flag: Origins, key?: string): string {
+		if (!key){
+			key = this.normalizePath(doc.fileName);
+		}
+		if (this.getSelectorType(key) === DocumentManager.Selectors.lsp){
+			if (this._cached.has(key) === false) {
+				const sources = DocumentSources.create(flag, doc);
+				this._cached.set(key, sources);
+			} else {
+				const sources = this._cached.get(key);
+				sources.native = doc;
+				this.tryUpdateInternal(sources);
+				sources.flags.add(flag);
+			}
+			return key;
+		} else {
+			return '';
+		}
+	}
+
+	getDocument(nDoc: vscode.TextDocument): ReadonlyDocument {
+		const key = this.documentConsumeOrValidate(nDoc, Origins.OPENED);			
+		return this._cached.get(key)?.internal;
+	}	
+
 	// Gets an array of PRJ ReadonlyDocument references, but verifies the content is based on a vscode.TextDocument's when available
 	private getProjectDocuments(): ReadonlyDocument[] {
 		const result: ReadonlyDocument[] = [];
 		if (ProjectTreeProvider.hasProjectOpened()) {
 			this.projectKeys.forEach(key => {
+				this.pathConsumeOrValidate(key, Origins.PROJECT);
 				if (this._cached.has(key)){
-					const sources = this._cached.get(key);
-					sources.flags.add(Origins.PROJECT);
-					if (sources.native && !sources.internal.equal(sources.native)){
-						sources.internal = ReadonlyDocument.getMemoryDocument(sources.native);
-					}
-					result.push(sources.internal);
-				} else {
-					const sources = DocumentSources.create(Origins.PROJECT, key);
-					this._cached.set(key, sources);
-					result.push(sources.internal);
+					result.push(this._cached.get(key).internal);
 				}
 			});
 		}
 		return result;
 	}
 
-
-	
 	// Gets an array of ReadonlyDocument references representing Origins.OPENED, but verifies the content is based on the native vscode.TextDocument
 	// When the internal & native documents match in content the existing version is returned, but will be regenerated if they are not
 	private getOpenedAsReadonlyDocuments(): ReadonlyDocument[] {
@@ -156,9 +171,7 @@ export class DocumentManager{
 		this.cacheKeys.forEach(key => {
 			const sources = this._cached.get(key);
 			if (sources.native && sources.flags.has(Origins.OPENED)){
-				if (!sources.internal || !sources.internal.equal(sources.native)){
-					sources.internal = ReadonlyDocument.getMemoryDocument(sources.native);
-				}
+				this.tryUpdateInternal(sources);
 				result.push(sources.internal);
 			}
 		});
@@ -171,9 +184,8 @@ export class DocumentManager{
 		this.cacheKeys.forEach(key => {
 			const sources = this._cached.get(key);
 			if (sources.flags.has(Origins.WSPACE)) {
-				if (sources.native && (!sources.internal || !sources.internal.equal(sources.native))){
-					sources.internal = ReadonlyDocument.getMemoryDocument(sources.native);
-				} else if (!sources.internal) {
+				this.tryUpdateInternal(sources);
+				if (!sources.internal) {
 					sources.internal = ReadonlyDocument.open(key);
 				}
 				result.push(sources.internal);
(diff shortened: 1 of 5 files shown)


extension/src/documents.ts
```diff
@@ -5,9 +5,37 @@ import { AutoLispExt } from './extension';
...
 
+enum Origins {
+	OPENED = 'O',
+	WSPACE = 'W',
+	PROJECT = 'P'
+}
+
+
+interface DocumentSources {
+	native: vscode.TextDocument;
```
hualin-adsk, line 16: we need to give its more descriptive name, I think `rawDoc` or any others

extension/src/documents.ts
```diff
@@ -5,9 +5,37 @@ import { AutoLispExt } from './extension';
...
+	internal: ReadonlyDocument;
+	flags: Set<Origins>;
```
hualin-adsk, line 18: It means that for one document, it can be opened, in workspace and in project files? maybe we need to add a comment here

extension/src/documents.ts
```diff
@@ -20,16 +48,63 @@ export class DocumentManager{
...
+		}
+		return result;
+	}
+
+	getDocument(nDoc: vscode.TextDocument) {
+		const key = nDoc.fileName.replace(/\//g, '\\');
+		if (this._cached.has(key)) {
+			const sources = this._cached.get(key);
+			sources.native = nDoc;
+			if (!sources.internal || !sources.internal.equal(sources.native)){
```
hualin-adsk, line 100: I think this is the change we check if the atoms forest should regenerate

extension/src/documents.ts
```diff
@@ -50,46 +125,66 @@ export class DocumentManager{
...
-			return ProjectTreeProvider.instance().projectNode.sourceFiles.map(x => 
-				this._opened.has(x.document.fileName) ? ReadonlyDocument.getMemoryDocument(this._opened.get(x.document.fileName)) : x.document
-				);
-		} else {
-			return [];
+			this.projectKeys.forEach(key => {
+				if (this._cached.has(key)){
+					const sources = this._cached.get(key);
+					sources.flags.add(Origins.PROJECT);
+					if (sources.native && !sources.internal.equal(sources.native)){
```
hualin-adsk, line 136: I think this 3 lines codes are duplicated several times, maybe extract to a function

extension/src/documents.ts
```diff
@@ -261,39 +413,37 @@ export class DocumentManager{
 		this.initialize();
 
 		AutoLispExt.Subscriptions.push(vscode.workspace.onDidCloseTextDocument((e: vscode.TextDocument) => {
-			if (this._opened.has(e.uri.fsPath) === true) {
-				this._opened.delete(e.uri.fsPath);
-			}
+			const source = this._cached.get(e.fileName.replace(/\//g, '\\'));
```
hualin-adsk, line 416: e.fileName.replace(/\//g, '\\') also calls several times, maybe extract to a util small function

hualin-adsk:
For F12 on a `setq`, it looks like it will find all the variables have the same name:
image

Maybe this is not an issue for F12, but it is risk for future F2 (rename)

extension/src/documents.ts
```diff
@@ -5,9 +5,37 @@ import { AutoLispExt } from './extension';
 import { ProjectTreeProvider  } from "./project/projectTree";
 
 
+enum Origins {
+	OPENED = 'O',
```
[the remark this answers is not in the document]
JD-Howard: These can go back to numbered enums now. I only added the strings to visually see which enum was getting added and purged during my tests. IE, I don't think we need these to have meaningful representations in the long term because they aren't exported.

extension/src/documents.ts
JD-Howard: In the context of this not being exported and living inside of `DocumentSources` interface, I actually think native and internal are meaningful. Clearly just an F2 to fix it if your comment isn't assuming use-cases beyond _documents.ts_ though.

extension/src/documents.ts
JD-Howard: Well, think about life cycle and the goal to recycle everything. 
Opening a file adds the OPENED flag, which is the only source of truth for what is actively opened. That file still could have had WSPACE & PRJ flags.
If you delete a file using file explorer while it was opened in the editor, the WSPACE flag gets removed, but we still have the reference with the OPENED flag and we should because VSCode doesn't close the document in that scenario.
If you are in a workspace, you open a document from file explorer, you then decide you need to close it, but add it to a PRJ, then nothing is lost.
If you open a PRJ, use F12, close that PRJ open a different one and later come back to it or open a different PRJ that also references that file, then nothing is lost.

extension/src/documents.ts
JD-Howard: Yes the `!sources.internal.equal(sources.native)` is the **cheap** check to see if there are disparities between our (internal) versions and the vscode (native) version of the document.

extension/src/documents.ts
JD-Howard: I know there is a lot of stuff that looks the same but sometimes has subtle differences. You are right, I should try a lot harder to clean this up with some meaningful function names even if they aren't recycled very much.

JD-Howard:
> For F12 on a `setq`, it looks like it will find all the variables have the same name:
It should only ever do this when it couldn't find anything in clear context. This is a result of a last ditch effort/fallback plan of locating all the first level setq's within the document root Sexpressions. 

Re: F2
When `IsFunction=False`, the F2 will have to do a much deeper mapping of a documents function calls and variable localization's to properly do work. This scenario will only happen on the local document. I am still mixed on whether I want F2 doing work across opened documents, PRJ's and workspace documents for function names.

extension/src/documents.ts
JD-Howard: I will change these to some variant of:
`this._cached.get(this.normalizePath(e.fileName))`

extension/src/documents.ts
JD-Howard: I made 2.5 functions that consume the vast majority of the complexity. Should be a lot clearer now. These primary handlers do some unnecessary things/checks for certain circumstance, but I don't believe this hurts anything.

extension/src/documents.ts
```diff
@@ -99,33 +206,44 @@ export class DocumentManager{
...
+				key = this.pathConsumeOrValidate(key, Origins.PROJECT);
+				if (key !== '') {
+					shouldExist.push(key);
 				}
 			});
+		}
+
+
+		/////////////////////// The extension appears to reset after every workspace change and this probably isn't really doing anything ///////////////////////
+		this.cacheKeys.filter(k => !shouldExist.includes(k)).forEach(obsolete => {
```
hualin-adsk, line 238: we also noticed this behavior, the extension will reset if open a workspace, open a folder.

When reset the extension, I think the `cacheKeys` will get lost (or reset) automatically?

extension/src/providers/gotoProvider.ts
```diff
@@ -8,7 +8,7 @@ import { SearchPatterns, SearchHandlers } from './providerShared';
 
 export class AutolispDefinitionProvider implements vscode.DefinitionProvider{
 	async provideDefinition(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken): Promise<vscode.Location | vscode.Location[]> {
-		const rDoc = ReadonlyDocument.getMemoryDocument(document);
+		const rDoc = AutoLispExt.Documents.getDocument(document);
 		let selected = '';
```
hualin-adsk, line 12: For the line 32, I still want to filter out to re-search in the `ActiveDocument`

hualin-adsk:
Thank you for refining this! 
I tried it in some real world cases:
 1. create project for all the `Express `tool in AutoCAD, F12 works well
 2. Open the lisp codes folder for AutoCAD Electrical which is written by lisp, it can work although the speed is slow for some large files.
I will try to use the `Chrome Dev tool `to see which part codes is the bottle neck codes.

extension/src/providers/gotoProvider.ts
JD-Howard: I think we'll ultimately wind up as much and more likely a lot more work filter that out than it currently would failing the simple pre-check.

JD-Howard:
I downloaded a bunch of random AutoLisp repo's and made a PRJ of the largest one. Think its largest file is just under 5000 lines. When I use F12 I get an infinite loop that never actually errors and once it gets going breakpoints don't function so it is really hard to find what is causing the problem. Who knows how exactly their code is randomly broken though.

hualin-adsk:
Yes, maybe the parser is the bottle neck. I will do some profiling, it is not this PR work. :)

hualin-adsk:
```
  (defun mode_tile_4_grp ( / )
    (cond
      ((= do_grp "1")
        (mode_tile "col_grp" 0)
      )
      (T (mode_tile "col_grp" 1))
    )
    (chk_accept_mode)
  )
```
If I select `mode_tile_4_grp` and hit F12, it will say there is no definition. I think this is a edge case.

hualin-adsk:
And we can associate this PR to this issue https://github.com/Autodesk-AutoCAD/AutoLispExt/issues/30

extension/src/providers/gotoProvider.ts
hualin-adsk: got it!

JD-Howard:
> And for the variables definition, maybe it only takes care the local variables (defined in function, or function parameter); don't search global variables (it means that for variable to search in the CURRENT file is enough).

We can do this for now until the F2 functionality is available. I fully expect the contextual mapping of a document to get pretty drastic in order to have any real confidence in rename operations.

>If I select mode_tile_4_grp and hit F12, it will say there is no definition. I think this is an edge case.

I think i need slightly more context. I pasted this function into a blank document, added `(mode_tile_4_grp)` after it and it was located just fine. Can you email a snippet of this that doesn't work please.

extension/src/documents.ts
JD-Howard: Yes, the only reason I left it was because this actually does do all the work to reset the state. However, when it is done, it also appears that it completely resets the extension wasting that work. If you don't have any ideas on how to persist that data past the reset, then it can go away.

extension/src/documents.ts
JD-Howard: So, did you come around on this or are we going with roDoc & vsDoc?

hualin-adsk: I am ok with the current name

extension/src/documents.ts
hualin-adsk: Thanks for explaining it!

extension/src/documents.ts
hualin-adsk: I don't have any ideal to persist the cache between the different folders. If the `cacheKeys` will be also reset automatically, I prefer to make the codes more simpler and cleaner (remove the `shouldExist`)

hualin-adsk:
`>> I pasted this function into a blank document, added (mode_tile_4_grp) after it and it was located just fine`
I mean that to select the function name after the `defun`, then hit F12, it can repro the issue.
Don't add the callings like `(mode_tile_4_grp) `

JD-Howard:
Your GIF illustrated an edge case problem in one of the new Sexpression functions. I fixed that and think everything else discussed was addressed. However, I am getting that infinite busy indicator on stuff I didn't previously have that problem with. So, I need to do some more investigating, but its certainly not going to be an easy thing to find without an actual error.

extension/src/providers/gotoProvider.ts
```diff
@@ -84,9 +84,9 @@ export class AutolispDefinitionProvider implements vscode.DefinitionProvider{
 			}
 		} while (flag);
 		// If we still haven't found anything check the 1st occurrence of setq's. This will find globals setqs and nested ones possibly inside other defuns
-		if (result.length === 0){
```
hualin-adsk, line 87: try this codes, and it will not break in the above loop:
```
(defun test (num) 
  (print num)
  (print "I am in the same line")
  (print "hello")
  (inner (* 20 5))
)
(defun inner (var) 
  (print var) ;; try F12 on this var
  (print "world")
)
(setq str (test 20)) 
```

hualin-adsk:
Try this simple case:
put this codes to a file and open that file, and it will not break in the above loop:

(defun test (num) 
  (print num)
  (print "I am in the same line")
  (print "hello")
  (inner (* 20 5))
)
(defun inner (var) 
  (print var) ;; try F12 on this var
  (print "world")
)
(setq str (test 20))

hualin-adsk:
I think it is introduced by this change: https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/49/commits/35afb0b8f1fde3608b70f1ba121d2b0ed99719fb#diff-604286706255a4a546d80ee16fd0a5263fae6e87f2a57f015b27870301f4858aL907

extension/src/providers/gotoProvider.ts
JD-Howard: Yeah, I am currently researching that. I am thinking I shouldn't have fixed the Sexpression, I should have checked the variable for null and acted accordingly.

hualin-adsk: It looks like the changes in Sexpression is to fix the edge case for (defun test ()), not for the variables

JD-Howard:
Global variables are working again, not scanning the whole document for Setq's and should now only be operating within the closest Defun of the the triggered position. Variables that are localized would never find a setq, they will always go up to the localization header. Does that sound like the behavior we want at this point? I expect this to change when the deep analysis for F2 comes in. 

IE, I would like to eventually make F12 to walk its way up to the Defun localization header; maybe even into another Defun that calls that function and also localizes that variable. Which I think will be easily possible after the F2 work is complete.

hualin-adsk:
It looks like working well. And (defun) can contain another (defun), I will try to test this point.

hualin-adsk:
Another issue:
```
 (setq lst (acet-geom-m-trans (list p1 p5 p6 p7 p1)
                     0 ;rk was na 6:28 PM 8/12/97
                     1
           );acet-geom-m-trans
 );setq
```
Select the comments `acet-geom-m-trans` and F12, it will report it can not find definition for `trans`

extension/src/providers/gotoProvider.ts
```diff
@@ -77,32 +77,36 @@ export class AutolispDefinitionProvider implements vscode.DefinitionProvider{
...
 			} else {
 				context = parent;
 			}
 		} while (flag);
 		// If we still haven't found anything check the 1st occurrence of setq's. This will find globals setqs and nested ones possibly inside other defuns
-		// if (result.length === 0){
-		// 	this.findInSetqs(doc, ucName).forEach(x => { result.push(x); });
-		// }
+		if (result.length === 0){
+			const possible = searchIn.findChildren(SearchPatterns.DEFINES, true).filter(p => p.contains(start));
```
hualin-adsk, line 88: I think this is the key codes to narrow the setq search

extension/src/providers/gotoProvider.ts
```diff
@@ -77,32 +77,36 @@ export class AutolispDefinitionProvider implements vscode.DefinitionProvider{
...
 
-	private findInSetqs(doc: ReadonlyDocument, ucName: string): vscode.Location[] {
+	private findInSetqs(sexp: Sexpression, ucName: string, fileName: string): vscode.Location[] {
 		const result: vscode.Location[] = [];
-		doc.findExpressions(SearchPatterns.ASSIGNS).forEach(setq => {
+		const found = sexp.findChildren(SearchPatterns.ASSIGNS, false);
+		found.forEach(setq => {
 			let isVar = false;
-			let cIndex = setq.nextKeyIndex(0);
+			let cIndex = setq.nextKeyIndex(0, true);
```
hualin-adsk, line 101: bool parameter hurts the codes readability.

hualin-adsk (approved):

Issue #40, opened 2020-10-27
Allow non-AutoLisp files in the project list.


morbius1st:
**Is your feature request related to a problem? Please describe.**
I work with more than just AutoLisp files for my project.  I use text files and .dat files but these are not permitted in the project list.

**Describe the solution you'd like**
1.  Allow the addition of any file in the project list.  
2.  provide a setting that provides a list of extensions to never add to the file list (e.g. .EXE files or .DLL files, etc.

**Describe alternatives you've considered**
none

**Additional context**
none

hualin-adsk:
it looks like old VLIDE project can not include other type files. To keep the back compatibility, it looks like we can not support this.
Maybe it needs to communicate widely if we decide to bring this feature.

JD-Howard:
It is not possible to support this and reverse compatibility. This needs to be closed.

rtogores:
Workspaces can include files of any kind. So why use projects?

hualin-adsk:
Autolisp has tons of legacy codes, and the old VLIDE command supports the projects file (to imagine that in real world users have hundreds of lisp files in different folders), it can be managed via PROJECT feature in old AutoCAD or vscode extension.

So there are two ways to work on AutoLisp in vscode:
1. Use workspace. I think this is more vscode flavor.
2. Use projects. This is legacy AutoCAD flavor.

hualin-adsk:
@LanDong-adsk This is as design, so I think it can be closed

Issue #47, opened 2020-11-10
Contextual menu item to add file to project


Rina-Y:
I'm just getting started trying out the .prj feature, I never used VLIDE for the lisp system I manage so I had never ran across it before.  There are hundreds of files in my lisp system and am trying to cherry pick specific files into subject related projects to try it out.  

The first thing I attempted when trying to add specific files to the project was right-click on the file in the workspace explorer and look for an item "Add to Current AutoLISP Project", hoping it was there.  Well, second thing, because the first thing was the "Add File to Project" button in the AutoLISP Project Manager area and added one manually that way.  I realized how much of a pain it would be to do this for the next 50 or so files, so I immediately *hoped* that there would be a contextual menu option for me to add things easily to the project.

hualin-adsk:
Maybe another solution is to support adding a folder to the project (automatically to add all the files in that folder to project).

JD-Howard:
I could be wrong, but I wouldn't think it would be very difficult to hook up a workspace "Add file to project" context menu item.

JD-Howard:
This has been implemented on the pending PR #51 
**When** the PR gets merged, the system is configured to auto-generate a [pre-release](https://github.com/Autodesk-AutoCAD/AutoLispExt/releases) installer. You could (if you want) download that and a VSCode extension that is designed to install *.vsix files if you want to beta test it; just don't expect documentation on pre-release features...

rtogores:
> 
> 
> Maybe another solution is to support adding a folder to the project (automatically to add all the files in that folder to project).

This is just as the Workspace feature works... Try the FILE > ADD FOLDER TO WORKSPACE menu option.

hualin-adsk:
@LanDong-adsk Or maybe this issue can be closed

LanDong-adsk:
The IDE supports multi-selection so it should be easy to add multiple files into the project.

Issue #38, opened 2020-10-27
Add "Select a folder" to add files to a project


morbius1st:
**Is your feature request related to a problem? Please describe.**
Not a problem - just makes things easier.

**Describe the solution you'd like**
All the selection of a folder and have the system select all of the files in the folder automatically.

**Describe alternatives you've considered**
none

**Additional context**
none

JD-Howard:
PR #51 will be adding folder support through a context menu on a workspace folder. 
However, the best method to add all the LSP's from a folder in the Project Tree View will probably just be using the filter search box within the file picker that already exists.

rtogores:
This will end up as Workspaces with a different name... ;-)

JD-Howard:
> This will end up as Workspaces with a different name... ;-)

AutoLisp doesn't have the ability to do explicit "import" declarations for function definitions/constants you expect to use from other files and often uses a lot of shared functionality. Projects help make those associations without creating a ton of false positives from your entire workspace as auto-completion values. Plus you must keep in mind that AutoCAD operates in the world of extreme reverse compatibility and this extensions is supposed to completely replace the original IDE at some point.

PR #51, opened 2020-11-21
Implements Enhancement #47

JD-Howard:
**Objective**
Should resolve the enhancement #47 request. Creates a context menu ability that adds Workspace files to an opened PRJ.

I was just going to look at this because I was curious. It turned out to be an exceptionally easy update and recycled all existing functionality/localizations. Copied existing `addFile2Project()` command, added the arguments related to workspace context menu event and added an optional argument to the existing function that added files to project. Which created this new `addWorkspaceFile2Project()` command with next to no effort.

**Tests Performed**
Successfully added files to an open PRJ from the workspace, verified the contextual aspects weren't available when not applicable and verified that we couldn't accidentally add non-LSP files. Also verified our original Treeview version still functioned.

### DIFF
--- a/extension/src/project/addFile2Project.ts
+++ b/extension/src/project/addFile2Project.ts
@@ -3,14 +3,14 @@ import { ProjectTreeProvider, isFileAlreadyInProject, hasFileWithSameName } from
 import { AutoLispExt } from '../extension';
 import * as path from 'path';
 
-export async function AddFile2Project() {
+export async function AddFile2Project(fileList?: vscode.Uri[]) {
     try {
         if (ProjectTreeProvider.hasProjectOpened() == false) {
             let msg = AutoLispExt.localize("autolispext.project.addfile.openproject", "A project must be open before you can add a file.");
             return Promise.reject(msg);
         }
 
-        let selectedFiles = await SelectLspFiles();
+        let selectedFiles = fileList ?? await SelectLspFiles();
         if (!selectedFiles)
             return; //user has cancelled the open file dialog
 
--- a/extension/src/project/addFile2Project.ts
+++ b/extension/src/project/addFile2Project.ts
@@ -90,17 +90,3 @@ async function SelectLspFiles() {
     if (fileUris && fileUris.length > 0)
         return Promise.resolve(fileUris);
 }
-
-
-export async function SelectFolder(): Promise<vscode.Uri[]> {
-    let label = AutoLispExt.localize("autolispext.project.addfolder.openlabel", "Add Folder");
-    const options: vscode.OpenDialogOptions = {
-        canSelectFolders: true,
-        openLabel: label
-    };
-
-    let fileUris = await vscode.window.showOpenDialog(options);
-    if (fileUris && fileUris.length > 0){
-        return Promise.resolve(fileUris);
-    }
-}
--- a/extension/src/project/addFile2Project.ts
+++ b/extension/src/project/addFile2Project.ts
@@ -90,3 +90,17 @@ async function SelectLspFiles() {
     if (fileUris && fileUris.length > 0)
         return Promise.resolve(fileUris);
 }
+
+
+export async function SelectFolder(): Promise<vscode.Uri[]> {
+    let label = AutoLispExt.localize("autolispext.project.addfolder.openlabel", "Add Folder");
+    const options: vscode.OpenDialogOptions = {
+        canSelectFolders: true,
+        openLabel: label
+    };
+
+    let fileUris = await vscode.window.showOpenDialog(options);
+    if (fileUris && fileUris.length > 0){
+        return Promise.resolve(fileUris);
+    }
+}
--- a/extension/src/project/projectCommands.ts
+++ b/extension/src/project/projectCommands.ts
@@ -3,7 +3,7 @@ import { OpenProject } from './openProject';
 import { ProjectTreeProvider } from './projectTree';
 import { openLspFile } from './openLspFile';
 import { IconUris } from './icons';
-import { AddFile2Project } from './addFile2Project';
+import { AddFile2Project, SelectFolder } from './addFile2Project';
 import { SaveProject, SaveAll } from './saveProject';
 import { excludeFromProject } from './excludeFile';
 import { getNewProjectFilePath, createProject } from './createProject';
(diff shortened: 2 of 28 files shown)


hualin-adsk:
maybe it looks like this issue https://github.com/Autodesk-AutoCAD/AutoLispExt/issues/38 can also be covered?

JD-Howard:
Yeah, that probably wouldn't be much more effort to do from the workspace. I wouldn't want to tack that onto this PR if we were talking about a new interface item within the Project Treeview though.

hualin-adsk:
This also reminds me that some users want a tree view in project panel, and also want to add some non lisp files to project files (I will check old VLIDE if can support the non-lisp files, and check how the vlx compiler to treat those non-lisp files).

JD-Howard:
The Treeview within the project panel is going to be a fairly large implementation change. The right click menu is currently filtered to only show up when an LSP file is selected and a PRJ is present. Removing half the filter isn't a big deal, but I'd have to do some research on the non-LSP file types. I know we have file type filtering at various locations, but I couldn't tell you what effects removing that could have.

hualin-adsk:
I tried to compile the codes, but didn't find the context menu. And How about to click the context menu if there is no project file opened/created?

JD-Howard:
If you have a project opened, then this should be working. You won't see the context menu items unless a PRJ is loaded. 

The last commit added the ability to process folders or even a mix of folders and files. If the folder contains non-LSP's, they were filtered out. I had to do this to work around the existing type checking I talked about earlier. IE, seems like our original implementation was specifically designed to abort the whole operation if it finds a non-LSP file.

hualin-adsk:
For the non-lisp files in project issue, I tested in legacy VLIDE, and it looks like the `FAS compiler` can not support such case:
```
(VLISP-PROJECT-LIST
  :name
  aaa
  :own-list
  ("ab" "a" "..\\sam.dwg")
  :fas-directory
  nil
  :tmp-directory
  nil
  :project-keys
  (:BUILD (:standard))
  :context-id
  :AUTOLISP
)
;;; EOF
```

JD-Howard:
I confirmed the same thing about the "other" file types. So, I would just close Issue #40 instead of creating a compatibility issue that probably won't add value.

hualin-adsk:
I see from workspace now it is easy to add to project:
addfilefromworkspace

And I think we need to add a prompt messages after finished to add the file to project, for examples `"10 lisp files have been added to test.prj project"`

JD-Howard:
Well, if we are going to deal with localizations on this, then maybe our workspace context menu should read "Add File(s) to Project"

I just recycled the old localization.

extension/src/project/projectCommands.ts
```diff
@@ -99,6 +100,45 @@ export function registerProjectCommands(context: vscode.ExtensionContext) {
             }
         }));
 
+        context.subscriptions.push(vscode.commands.registerCommand('autolisp.addWorkspaceFile2Project', async (clickedFile: vscode.Uri, selectedFiles: vscode.Uri[]) => {
```
hualin-adsk, line 103: And I see in the project root node to right click there is also an menu "add file to project", for this menu it is better that it can also support to select a folder.

extension/src/project/projectCommands.ts
```diff
@@ -99,6 +100,45 @@ export function registerProjectCommands(context: vscode.ExtensionContext) {
...
+            }
+            let selectedDirs = selectedFiles.filter(f => fs.statSync(f.fsPath, { bigint: false}).isDirectory());
+            selectedFiles = selectedFiles.filter(f => fs.statSync(f.fsPath, { bigint: false}).isFile());
+            
+            selectedDirs.forEach(dir => {
+                fs.readdirSync(dir.fsPath).forEach(name => {
+                    const fspath = dir.fsPath.replace(/[\/\\]$/, '') + '\\' + name;
+                    if (fs.existsSync(fspath) && fs.statSync(fspath, { bigint: false}).isFile()) {
+                        if (AutoLispExt.Documents.getSelectorType(fspath) === AutoLispExt.Selectors.lsp){
+                            selectedFiles.push(vscode.Uri.file(fspath));
```
hualin-adsk, line 115: Maybe it needs to test if the selectedFiles is null (the workspace has no lisp files)

extension/src/project/addFile2Project.ts
```diff
@@ -3,14 +3,14 @@ import { ProjectTreeProvider, isFileAlreadyInProject, hasFileWithSameName } from
 import { AutoLispExt } from '../extension';
 import * as path from 'path';
 
-export async function AddFile2Project() {
+export async function AddFile2Project(fileList?: vscode.Uri[]) {
```
hualin-adsk, line 6: the typescript new syntax is nice for null case

JD-Howard: Yeah, the changes to this file do really show the value of the Typescript update. I didn't have to add any additional lines to recycle the method!

extension/src/project/projectCommands.ts
JD-Howard: You don't think the ability to use standard multi-selection is sufficient? Does VSCode even have a decent folder picker? I feel like that's a pretty large deficiency existing in all newer Microsoft related frameworks.

extension/src/project/projectCommands.ts
JD-Howard: I don't think this is even possible. From what I can tell, the context menu won't show up unless its on/over an LSP file or a Directory. If it is over a directory, then we have at least 1 URI even if it doesn't contain any files. IE, it just won't have anything to do work with.

package.json
```diff
@@ -590,6 +599,18 @@
...
 				}
+			],
+			"explorer/context": [
+				{
+					"when": "resourceLangId == autolisp && autolisp.hasProject",					
+					"command": "autolisp.addWorkspaceFile2Project",
+					"group": "navigation@1"
+				},
+				{
+					"when": "explorerResourceIsFolder && autolisp.hasProject",
```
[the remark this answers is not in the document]
JD-Howard: This is, but I'll draw your attention Line 99 simply because it is where I recycled the previous localization that I am questioning should change to possibly pluralize the localized statement.

extension/src/project/projectCommands.ts
```diff
@@ -99,6 +100,45 @@ export function registerProjectCommands(context: vscode.ExtensionContext) {
             }
         }));
 
+        context.subscriptions.push(vscode.commands.registerCommand('autolisp.addWorkspaceFile2Project', async (clickedFile: vscode.Uri, selectedFiles: vscode.Uri[]) => {
+            if (!selectedFiles || selectedFiles.length === 0 || getWarnIsSearching()){
+                return;
+            }
+            let selectedDirs = selectedFiles.filter(f => fs.statSync(f.fsPath, { bigint: false}).isDirectory());
+            selectedFiles = selectedFiles.filter(f => fs.statSync(f.fsPath, { bigint: false}).isFile());
```
hualin-adsk, line 108: Your codes cthere support to select a dir. But in the project node it doesn't support to select dir. I think to keep consisstene in hte project node context menu it is best to have this.

JD-Howard: Okay, but I am not very good at designing icons. So, which SVG's would you like associated with this new interface element?

hualin-adsk: I only mean to refine the functionality of the menu item to support selecting dir, see pic:
addfilefromworkspace-menu

extension/src/project/addFile2Project.ts
```diff
@@ -90,3 +90,17 @@ async function SelectLspFiles() {
     if (fileUris && fileUris.length > 0)
         return Promise.resolve(fileUris);
 }
+
+
+export async function SelectFolder(): Promise<vscode.Uri[]> {
+    let label = AutoLispExt.localize("autolispext.project.addfolder.openlabel", "Add Folder");
```
hualin-adsk, line 96: Thanks for your patience!

Sorry I didn't make me understood on this item. 
Please see the GIF: 
https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/51#discussion_r528318730

And I mean that we only improve the "add file to project" menu item and make it support to select a folder.

Rina-Y:
Does this work for individual files or multiple files selected inside the workspace rather than just folders? Was just curious due to the watching your screencap that it only had folders being added.

JD-Howard:
> 
> 
> Does this work for individual files or multiple files selected inside the workspace rather than just folders? Was just curious due to the watching your screencap that it only had folders being added.

If your in a workspace, and you Shift+Select multiple files, all the files/folders in that selection would get added to the project. Note that this will not recursively dig through sub-folders; just the top level files.

hualin-adsk (approved):

hualin-adsk:
It is best that adding a prompt message for how many files have been added to project file.

JD-Howard:
I just put the quantity message back in from commit before the revert. I also added File Type filtering for *.LSP's since we've now determined PRJ's only want to work with LSP's and a workspace presents a lot of opportunity for non-LSP's to get in the way.

PR #52, opened 2020-11-23
AtomsForest Replacement & Tests

JD-Howard:
**Objective**
It was discovered that the existing AtomsForest in the LispParser has an exponential performance problem. On really large documents (8000 lines) this was taking a very substantial amount of time to generate. This PR intends to replace that functionality for (at least) new features suffering from those discovered performance impacts.

**Abstraction**
The `getDocumentSexpression()` function was tacked onto the existing LispParser object namespace to isolate it from the existing code base and was designed to parse the contents of a document or string into an Sexpression. 

**Implementation**
This takes a much more linear approach of reducing document content into LispAtom & Sexpression segments. It does this by parsing the current level LispAtoms and then recursively calling itself to generate child Sexpressions. While persisting its index/line tracking mechanism object into those recursive calls.

**Tests Performed**
New tests *.TS was added to the test suite. I also took a massive (3500 line) document, filtered out the Sexpressions from both the OLD/NEW parsers and compared the results for anomalies. Which generated the **Quirks** section below.

**Additional Information**
While doing this and comparing the output to the existing atomsForest, a few **Quirks** were discovered in the original versions column # values. We have a very fundamental dependency on Sexpression's that needs to be **heavily** researched to check the viability of replacing **any** of the existing mechanisms with this new version. This will inevitably be a long running and potentially very expansive PR if we determine all existing Sexpression code needs to be normalized into this version.

**Quirks**
Note: In all situations, except where stated below, the column index is equal to the start of a LispAtom or Sexpression.
- Quote Marks Prefixing a Parenthesis
  - Old Version
    - The Sexpression Parent, Quote & the Parenthesis all have the same column index. Which all literally represent the location of the parenthesis within the document.
  - New Version
    - The Sexpression Parent & Quote have the same column index and literally represent their location within the document. The parenthesis also has a column index that literally represent its location in the document.

- Closing Parenthesis
  - Old Version
    - The closing parenthesis LispAtom column index represents the true end of the Sexpression.
  - New Version
    - The closing parenthesis LispAtom column index represents the start of that LispAtom.


extension/src/format/parser.ts
```diff
@@ -182,4 +182,149 @@ export class LispParser {
...
+        result.line = tracker.line;
+        result.column = tracker.column;
+        result.linefeed = tracker.linefeed;
+        
+        let isAuthorized = true;
+        let state = ParseState.UNKNOWN;
+        let grpStart: vscode.Position = null;
+        let temp = '';
+        
+        while (tracker.idx < documentText.length && isAuthorized) {
```
hualin-adsk, line 227: I think this is duplicate with the current LispParser, and you plan to replace the current LispParser with this new one? We should keep the codes consistency, to follow the [DRY](https://dzone.com/articles/software-design-principles-dry-and-kiss) rule.

JD-Howard: If you want me to do all this in the shadows and only bring it out after I can replace the whole thing, then I will..... but I thought this was too fundamental to operate in that way.

hualin-adsk: It looks like the old lisp parser has 10x slow down performance issue than this one you written. I will take some time to do profiling. 
If we decided to replace the old lisp parser, it needs to do lots of regression tests for format (document format, selection format and the enter key format, and the project file format).
If you also think that it is large task, I agree that we hold on this PR, and focus on other tasks. (And this is fundamental, I think that it is worthy to take more time to refine/improve it).

JD-Howard: I am of the opinion that (yes) this is very similar, but (no) it provides entirely different use cases from the `LispParser.AtomsForest`

I say that because of an attempt to use this because the `LispFormatter.formatGut()` does appear to be dependent on raw string values that the AtomsForest provides and I have been saying for several PR's I need something with LispAtom's not strings.

hualin-adsk:
I want to hold on this PR for sometime, and do some profiling on the legacy parser. I don't want to keep two simialr codes in the code base.
For the profiling I will do it, and sync up the status in this PR latter.

JD-Howard:
Do you want this closed or left here for reference?

FYI, I won't be doing anything with this or Sexpressions until I know what is going on. Ultimately, I need a way to get the `Array<LispAtom|Sexpression>` context. I don't really care how that happens, but it is something I need for future developments.

Also note, I was able (locally) to easily modify my version to accept a starting `offset` much like the `tokenizeString` uses.

hualin-adsk:
Thank you! I want to left this PR there for references.

JD-Howard:
Added my revisions that gives it index initialization. Just getting it off my plate. I need to go work on an edX final project for a bit.

JD-Howard:
Didn't get all of them, but removing all the Imports and object references from _vscode-languageclient_ did remove the large block of errors. I think this might be due to our JSON configs handling many of the things that namespace was intended to cover, but that's just a guess...

hualin-adsk:
I am trying to test the branch in real world large project with several huge lisp files(10000+ line codes). It can improve the performance, but it looks like the improvement is not so big expected.

JD-Howard:
Are you testing this by running the extension or through the test suite? Right now it was only configured to the test suite, you won't see any differences while actively running the extension.

IE, my form of testing involved pasting mastic chunks of lisp code into the `/test_case/pdfMakrups.lsp` but without saving those changes to the PR.

hualin-adsk:
Sorry I was trying it by running the extension, missed that it didn't hook to extension!

hualin-adsk:
Could you point it to me how to enable this PR when hit F12? I want to do some tests to check if we can merge this PR to version 1.4.0

JD-Howard:
I was thinking that the ReadOnlyDoucment object would changes its AtomsForest reference to return the values from this function. I was the one who embedded the AtomsForest in the ROD, so the only things in the extension dependent on that value are "new" features.

This PR would be inert if it were to get merged today. More work still needs to be done to implement it and I would need to go unravel some things that were added to the Sexpression to support prior PR's using the old AtomsForest.

JD-Howard:
Closing this because it was updated to operate F12 in a new PR that was derived from the most current commit chain

PR #55, opened 2020-12-09
Deprecate to use the launch.json to debug autolisp files

hualin-adsk:
#### Objectives
In this commit I trid to remove the support for the launchjson.

Now the debug settings (attach and launch for acad installation path) is in the extension "settings" page, which is a global settings. So it has no need to use the launch.json file. In fact the file has been deprecated in the version 1.3.0, and it only reads the settings (can not set the settings).

In this version I tried to remove the reading setting support for launch.json, it has some issues to debug the lisp after upgrade the vscode. Some customers reported that after updating vscode, it may have issue to attach debugging autocad. 

#### tests performed
tried to open a folder with the legacy launch.json, and do following settings:
 1. attach debug
 2. launch debug
 Then do some changes in the launch.json file, and do above tests again

#### Screen Shot
No new behavior introduced, not applicable


extension/src/debug.ts
```diff
@@ -125,31 +117,22 @@ class LispLaunchConfigurationProvider implements vscode.DebugConfigurationProvid
...
+             newConfig.program = vscode.window.activeTextEditor.document.fileName;
 
-        if (config["type"] === launchCfgType) {
+        if (newConfig["type"] === launchCfgType) {
             // 1. get acad and adapter path
-            //2. get acadRoot path
-            //2.1 get acadRoot path from launch.json
-
+            // 2. get acadRoot path
             let productPath = "";
```
Sen-real, line 132: Can we just:
`let productPath = getExtensionSettingString(LAUNCH_PROC);`
If this API returns "", then there should be no difference.

extension/src/debug.ts
```diff
@@ -184,16 +168,12 @@ class LispLaunchConfigurationProvider implements vscode.DebugConfigurationProvid
...
                 let params = "";
-                if (config["attributes"]) {
-                    params = config["attributes"]["params"] ? config["attributes"]["params"] : "";
-                }
-                else {
-                    let text = getExtensionSettingString(LAUNCH_PARM);
-                    if (text)
-                        params = text;
-                }
+                let text = getExtensionSettingString(LAUNCH_PARM);
```
Sen-real, line 171: similarly, if getExtensionSettingString returns "", we can just
`ProcessPathCache.globalParameter = getExtensionSettingString(LAUNCH_PARM);`

Sen-real (approved):

JD-Howard:
I didn't clone-test this, but nothing in here looks wrong or even subjective to me. Cleaner and more succinct.

JD-Howard (approved):

PR #63, opened 2021-01-12
make correct package for vsix file in Github CI

hualin-adsk:
##### Objective
Github CI didn't package the visx file correctly, it needs to call the pack.py to copy some utility files to correct location.

##### Abstractions
In the pack.py it runs the `npm install`, and copy some utilities files for example the search tool exe files to the correct locations, then `gulp package `can make an expected vsix file for us. So it means that the several pre-release packages we made in CI are not correct. (I will do house keeping latter).

I also make the gulp to install it locally instead of globally, because CI machine to complains that it has no access to some global path.

##### Tests performed
Checked the PR preflight messages that there is no warning to run the pack.py.

##### Screen shot
Not applicable for this PR

### DIFF
--- a/pack.py
+++ b/pack.py
@@ -11,8 +11,8 @@ import sys
 
 def init():
     print("===============================================")
-    print("             try to install gulp-cli globally")
-    os.system("npm install --global gulp-cli") # nosec
+    print("             try to install gulp-cli")
+    os.system("npm install gulp-cli") # nosec
 
     os.system("npm install --unsafe-perm") # nosec
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,4 +8,4 @@
 
 
 ##### Screen shot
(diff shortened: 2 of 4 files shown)


pack.py
```diff
@@ -11,8 +11,8 @@
 
 def init():
     print("===============================================")
-    print("             try to install gulp-cli globally")
-    os.system("npm install --global gulp-cli") # nosec
+    print("             try to install gulp-cli")
+    os.system("npm install gulp-cli") # nosec
```
JD-Howard, line 15: So, we don't need to have gulp-cli installed globally?

hualin-adsk: yes, I think it doesn't need to do that, otherwise it will have access deny issue in the CI machine.
To change this I see the CI is good:
https://github.com/Autodesk-AutoCAD/AutoLispExt/runs/1687520696

Sen-real (approved):

PR #64, opened 2021-01-13
fix the CI to publish vsix package

hualin-adsk:
##### Objective
It looks like after this PR https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/63 the CI complains that it can not find the vsix file to do publish:
 Error: ENOENT: no such file or directory, stat './autolispext-1.3.5-da09c7ad1cc9499079d091fc510a798476d8f805.0.vsix'

More details messages see CI page: https://github.com/Autodesk-AutoCAD/AutoLispExt/runs/1692115357

##### Abstractions
To fix the issue I tried to change the main.yml file to publish the `autolispext.vsix` instead of the one with the commit ID package.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,6 +69,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }} # This pulls from the CREATE RELEASE step above, referencing it's ID to get its outputs object, which include a `upload_url`. See this blog post for more info: https://jasonet.co/posts/new-features-of-github-actions/#passing-data-to-future-steps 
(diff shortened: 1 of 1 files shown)


Sen-real:
Where is the file published to? is it allowed to publish a unique vsix for each build? if so, perhaps we should append version+build number in the file name as we did in the internal CI

hualin-adsk:
I think the upload path is related to commit id: https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/64/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R71

@szilvaa-adsk Do you have some ideal one this? I am confused why the old "gulp package" can assign the commit id to the result package name.

Sen-real (approved):

PR #67, opened 2021-01-18
refine CHANGELOG and fix online help bug

hualin-adsk:
##### Objective
In this commit I want to fix the online help bug, because the latest AutoCAD has not been released, so the ACAD2022 help can not get. And I also refined the changelog text.

##### Abstractions
Our tester complains these two issues to us that the "online help" is broken at this point that the new AutoCAD 2022 has not been released. And the change log text is also not correct. So in this commit I aimed to resolve those two issues.

##### Tests performed
- to test the "Online help" can work well for built in keywords

##### Screen shot
Not applicable for this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
-## [1.4.0] - 2020-09-21
+## [1.4.0] - 2021-01-20
 a. Added Right Click -> 'Open Online Help' feature. This makes it possible for users to quickly access official documentation on a selected symbol.
+b. Support to insert custom regions.
+c. F12 "Go to definition" support.
+d. Fix DCL comments issue and other minor bugs.
 
-b. Online Help GUID targets were provided by a new data source (webHelpAbstraction.json) that will likely be used to provide future intellisense enhancements
-
-c. Fixed a bug with DCL comments. Both single line and Block comments should now be working.
-
+## [1.3.2] - 2020-12-17
+Remove the call to the deprecated vscode APIs
 
 ## [1.3.0] - 2020-06-12
 a.	Support AutoLISP project editing. Support find and replace in the open project.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
-## [1.4.0] - 2020-09-21
+## [1.4.0] - 2021-01-20
```
hualin-adsk, line 1: This time will be refined when it is time to release 1.4.0

extension/src/help/openWebHelp.ts
```diff
@@ -148,7 +148,7 @@ class WebHelpEntity {
 
 	static getDefaultHelpLink(): string {
 		let lang: string = WebHelpEntity.getLanguageUrlDomain();
-		let year: number = new Date().getFullYear() + 1;
+		let year: number = new Date().getFullYear();
```
Sen-real, line 151: Can we have a fallback logic here? if 2022 doesn't exist, then we visit 2021?

JD-Howard: Its currently a blind open of a URL that is expected to exist. If we want to test it, then we'll have to do all kinds of HTTP GET work. Where does that end? If Current Year & CY + 1 both fail do we start trying other languages too?

In an email Hualin and I drew the conclusion it made sense to just target the current year to forever avoid this problem. It's not like the help really changes much from year to year and our JSON won't have the GUID's for the next releases new functions without doing the HelpAbstractionGenerator.exe process. If it next release doesn't exist, then we couldn't perform that collection step anyway. 

With that said, we could probably embed the target year into the webHelpAbstraction.json, but that is my only reasonable way to solve this dilemma without doing work that won't offer real benefits. If you want to go down that road, merge this PR and I can work on the various things associated with the C# project in a different one.

Sen-real (approved):

PR #68, opened 2021-01-19
Help abstraction update

JD-Howard:
##### Objective
Embed the year reference within the webHelpAbstraction.json export.

##### Abstractions
Removed all hard coded and/or dynamically generated year paths related to documentation. Implemented a user prompt during abstraction creation, the provided value becomes the year baseline in all places from that point forward.

##### Tests performed
**C#**
- Verified the year would be dynamically added to existing json exports
- Verified the pages wouldn't load if an unreleased year was provided

**TS**
- Verified the help link functionality still works and used the json year value


extension/src/help/webHelpAbstraction.json
```diff
@@ -1,4 +1,5 @@
 {
+  "year": "2021",
```
hualin-adsk, line 2: So now this is a hardcode year now?

JD-Howard: Not entirely,
It now represents "what was available" and theoretically "when" the webHelpAbstraction.json was last updated. I did build in a "fallback" of 2021 in the C#.

utils/HelpAbstractionGenerator/StaticUtilities.cs
```diff
@@ -38,13 +38,18 @@ public static string wrapContentIntoHtmlDocument(List<HAP.HtmlNode> bodyContent)
...
         public static string getHelpURL(string guid)
         {
             return getDefaultHelpURL() + "?guid=GUID-" + guid;
         }
         public static string getDefaultHelpURL()
         {
-            return "https://help.autodesk.com/view/OARX/2021/" + System.Globalization.CultureInfo.CurrentCulture.ThreeLetterWindowsLanguageName + "/";
+            if (year == "")
+                year = Microsoft.VisualBasic.Interaction.InputBox("This should be the most current version of the Autodesk documentation and may not be the current year shown", "Enter 4 digit year", DateTime.Now.Year.ToString()).Trim();
+            if (year == "")
```
Sen-real, line 50: how about string.IsNullOrWhiteSpace(year)

Sen-real (approved):

utils/HelpAbstractionGenerator/StaticUtilities.cs
JD-Howard: It can't hurt, but I would note that line 41 is its definition, it gets initialized to an empty string and is only ever set inside  this getDefaultHelpURL() function. So, its technically fine as is.

PR #71, opened 2021-01-27
refine the changelog text and highlight nil

hualin-adsk:
##### Objective
To prepare releasing the version 1.4.0, I want to update the change log text reviewed by our User Assistance.

##### Abstractions
I also fix a TODO item in the codes for localization missing. After fixed it in CI you may see some gulp complaints. This is expected, and it will be fixed after I updated the localization strings:
```
[17:20:04] [i18n] Generating localized messages for 'ptb' resulted in the following problems:
[17:20:04] [i18n]
[17:20:04] [i18n] No localized message found for key autolispext.regionsnippet.title
[17:20:04] [i18n] No localized message found for key autolispext.project.closeproject.title
[17:20:04] [i18n]
```
I also add the nil to the highlight keyword.

##### Tests performed
 - markdown preview the change log text
 - highlight the nil text

##### Screen shot

image


JD-Howard (approved):
Looks good to me. Double checked the regex in RegExr just because the double escapes were confusing me.
Your change log could add the ability to close Projects. IE, it does have an effect on the F12 and eventually the auto-completion.

PR #91, opened 2021-02-24
Analytical Parser

JD-Howard:
**Objective**
It was discovered that the existing AtomsForest in the LispParser has an exponential performance problem. On really large documents (8000 lines) this was taking a very substantial amount of time to generate. This PR intends to replace that functionality for (at least) new features suffering from those discovered performance impacts. The existing Parser will continue to be used for its original purpose (formatting) since it is dependent on knowledge of white space and this has a focus on concrete Atoms & Sexpressions.

I tweaked the launch.json & tasks.json files to compile the TS from F5 so I didn't have to keep running other scripts as a secondary step. From a fresh clone you would still need to use the pack.py to get the file dependencies copied.

**Abstraction**
The `getDocumentSexpression()` function was tacked onto the existing LispParser object namespace to isolate it from the existing code base and was designed to parse the contents of a document or string into an Sexpression. 

**Implementation**
This takes a much more linear approach of reducing document content into LispAtom & Sexpression segments. It does this by parsing the current level LispAtoms and then recursively calling itself to generate child Sexpressions. While persisting its index/line tracking mechanism object into those recursive calls.

**Tests Performed**
New tests *.TS was added to the test suite. I also took a massive (3500 line) document, filtered out the Sexpressions from both the OLD/NEW parsers and compared the results for anomalies. Which generated the **Quirks** section below. Mild alteration to the `ReadOnlyDocument` object to house a document wide Sexpression and to change the `AtomsForest` property to `Array<LispAtom|Sexpression>`. This effectively hooked up F12 to this new data model.

**Additional Information**
While doing this and comparing the output to the existing atomsForest, a few **Quirks** were discovered in the original versions column # values. 

**Quirks**
Note: In all situations, except where stated below, the column index is equal to the start of a LispAtom or Sexpression.
- Quote Marks Prefixing a Parenthesis
  - Old Version
    - The Sexpression Parent, Quote & the Parenthesis all have the same column index. Which all literally represent the location of the parenthesis within the document.
  - New Version
    - The Sexpression Parent & Quote have the same column index and literally represent their location within the document. The parenthesis also has a column index that literally represent its own location in the document.

- Closing Parenthesis
  - Old Version
    - The closing parenthesis LispAtom column index represents the true end of the Sexpression.
  - New Version
    - The closing parenthesis LispAtom column index represents the start of that LispAtom.
      - Note, Sexpression.getRange() was updated to function with the new version. Nothing in the "formatting parser" model was using this and instead made range specific calculations manually. It should be safe to consider the `Sexpression.getRange()` part of the "analytical parser" model that does not intend to do anything like that manually, but rather lean on providers such as getRange.

### DIFF
--- a/extension/src/project/readOnlyDocument.ts
+++ b/extension/src/project/readOnlyDocument.ts
@@ -106,13 +106,13 @@ export class ReadonlyDocument implements vscode.TextDocument {
 
     
     // Converted this from a constant data feature into an on-demand feature that once used is essentially cached for future queries.
-    get atomsForest(): Array<string|Sexpression> {
+    get atomsForest(): Array<LispAtom|Sexpression> {
         if (this.languageId === DocumentManager.Selectors.lsp) {
-            if (this._atomsForest){
-                return this._atomsForest;
+            if (this._docExpression){
+                return this._docExpression.atoms;
             } else {
                 this.updateAtomsForest();
-                return this._atomsForest;
+                return this._docExpression.atoms;
             }
         } else {
             return [];
@@ -127,20 +127,14 @@ export class ReadonlyDocument implements vscode.TextDocument {
             if (content) {
                 this.initialize(content, this.languageId);
             }
-            let parser = new LispParser(this);
-            try {
-                parser.tokenizeString(this.fileContent, 0);        
-                this._atomsForest = [...parser.atomsForest];    
-            } finally {
-                parser = undefined;
-            }
+            this._docExpression = LispParser.getDocumentSexpression(this.fileContent);  
         }
     }
 
     fileContent: string;
     lines: string[];
     eolLength: number;
-    private _atomsForest: Array<string|Sexpression>; // Added to drastically reduces complexity in other places.
+    private _docExpression: Sexpression; // Added to drastically reduces complexity in other places.
 
     //#region implementing vscode.TextDocument
 
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,6 +9,7 @@
             "name": "Extension Client",
             "type": "extensionHost",
             "request": "launch",
+            "preLaunchTask": "buildproject",
             "runtimeExecutable": "${execPath}",
 			"args": ["--extensionDevelopmentPath=${workspaceRoot}"],
 			"outFiles": ["${workspaceRoot}/extension/out/**/*.js"],
@@ -18,6 +19,7 @@
             "name": "Extension Tests",
             "type": "extensionHost",
             "request": "launch",
+            "preLaunchTask": "buildproject",
             "runtimeExecutable": "${execPath}",
             "args": [
                 "--extensionDevelopmentPath=${workspaceFolder}",
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -4,6 +4,7 @@
     "version": "2.0.0",
     "tasks": [
         {
+			"label": "buildproject",
 			"type": "npm",
 			"script": "compile",
 			"group": "build",
--- a/extension/src/format/parser.ts
+++ b/extension/src/format/parser.ts
@@ -1,6 +1,6 @@
 import * as vscode from 'vscode';
 import { ListReader, CursorPosition } from "./listreader";
-import { Sexpression } from "./sexpression";
+import { LispAtom, Sexpression } from "./sexpression";
 
 class LeftParentItem {
     public location: Number;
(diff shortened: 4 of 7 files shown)


extension/src/project/readOnlyDocument.ts
```diff
@@ -127,20 +127,14 @@ export class ReadonlyDocument implements vscode.TextDocument {
...
                 this.initialize(content, this.languageId);
             }
-            let parser = new LispParser(this);
-            try {
-                parser.tokenizeString(this.fileContent, 0);        
-                this._atomsForest = [...parser.atomsForest];    
-            } finally {
-                parser = undefined;
-            }
+            this._docExpression = LispParser.getDocumentSexpression(this.fileContent);
```
Sen-real, line 130: Please note that the ReadOnlyDocument is used for project Open&Save.
I suggest that we limit the use of the new parser to new features for now, and gradually use it to replace the old one after well tested.

JD-Howard: The `_docExpression` is only currently used internally by the ReadOnlyDocument's `get atomsForest()` property. So, both have only ever been used to power new features. I agree on the testing, I wouldn't be the least bit surprised if malformed syntax will cause some some edge case anomalies. I don't really expect those to surface in clear ways until AutoCompletion starts attempting to use this.

Sexpression differences are still the one thing I am most concerned about. Should we add a property to the Sexpression class that can identify whether it is part of the analytical model? We wouldn't have to update any other code, simply initialize the value and this new stuff could explicitly set it to something unique. I could then run through all the "newer" Sexpression functions putting their current work in the "new" code path and leave ToDo notes on the else. At least then it would create some clarity on where that divide happened and what was built for which task.

Sen-real: I see, you're right.

Sen-real (approved):

extension/src/project/readOnlyDocument.ts
JD-Howard: Another thought on the Sexpressions delima:
Maybe we should create a class that inherits Sexpression or that Sexpression inherits to completely divide what is part of the original formatting context and what is going to be used for responsive analytical work? If you think about it, the Sexpression currently has a ton of formatting specific behaviors that have nothing to do with the core of what an Sexpression really is; which is a container....

Sen-real: Yes, it makes great sense.

JD-Howard:
I have Sexpression and SexpressionBase divided out with the Sexpression deriving from the SexpressionBase because that was the best way to protect all the existing code. The 1 caveat is that I wanted to move LispAtom into SexpressionBase because structurally makes sense to go with the more fundamental SexpressionBase. This is going to cause 5 documents to get modified to add that import and 3 more for the actual Sexpression references. **Should that be in a different PR?
Other Notes:**
- the current PR involves 3 of those 5. 
- I did keep all code from Both the Sexpression and the new SexpressionBase together and commented out what should be moved to make 1-1 comparison easier

Sen-real:
> I have Sexpression and SexpressionBase divided out with the Sexpression deriving from the SexpressionBase because that was the best way to protect all the existing code. The 1 caveat is that I wanted to move LispAtom into SexpressionBase because structurally makes sense to go with the more fundamental SexpressionBase. This is going to cause 5 documents to get modified to add that import and 3 more for the actual Sexpression references. **Should that be in a different PR? Other Notes:**
> 
> * the current PR involves 3 of those 5.
> * I did keep all code from Both the Sexpression and the new SexpressionBase together and commented out what should be moved to make 1-1 comparison easier

@hualin-adsk what do you think? for me I would suggest to keep the current code in PR91.

JD-Howard:
Okay, hopefully we don't have to change much in this one since they are both effecting the same files. I'll get a PR created a bit later, but I have moved the work to a different branch and pushed it to my GitHub if you want to look at it.
https://github.com/Autodesk-AutoCAD/AutoLispExt/compare/main...JD-Howard:SexpressionBase

hualin-adsk:
I am ok with the current implementation, and it can improve some performance for F12.
I think for the code refractoring it is not an urgent thing. If in the [third time](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) we find it need to do code refine, we do it.

I think we need to take care some user feedback for the F12 feature.

JD-Howard:
Thanks for the insight Hualin, I haven't seen that rule before. I believe separating the formatting context from the container context on the Sexpression will probably make doing that a lot easier. I also intend on approaching formatting at some point to create a "classic" version of formatting that most users would be more familiar with. So, that would probably be a good time to see how viable a merge would be.

I did read the F12 issue and it seemed like it was "working as intended" to me. I'll go read it again or ask some questions to see if I can understand the desire better.

PR #106, opened 2021-03-22
To report code coverage of tests run with "npm run test"

Sen-real:
##### Objective
To report code coverage of "npm run test"

##### Abstractions
Tried several different options, and it's finally done following steps in this article:
https://frenya.net/blog/vscode-extension-code-coverage-nyc

##### Tests performed
A "coverage" folder is generated with code coverage report in HTML:
image
image

Also print coverage summary in console so that we could easily check the result on the CI in future:
image

##### Screen shot
See "Tests performed".


hualin-adsk:
This is very cool!

Sen-real:
> This is very cool!

Thanks, this is so quick! let me add you as a reviewer :)

szilvaa (approved):

szilvaa-adsk (approved):

LanDong-adsk:
I would like to report coverage if an option is specified, for example, --coverage. In CI, I don't think we care about coverage so we should be able to control when to generate coverage.

PR #111, opened 2021-03-26
Scan only out folder for codeql

nigma456:
##### Objective
Specify only /out/ folder to do the codeql since I found it takes too long time for codeql on Github CI.


LanDong-adsk:
we can't read the warnings on output js file. We need to see what is the error with ts file.

PR #115, opened 2021-03-27
Fix test failure on Mac about end-of-line

Sen-real:
##### Objective
A potential bug is discovered by running automation tests on Mac. Error "failed to read string on top level" is printed in console.

##### Abstractions
It's caused by an error in a line of code. Unlike Windows, when running tests on Mac, the calls to split and join methods (which was added for an edge case, and not needed with the given test data) will change the original string, and this is why the error is discovered. On Windows, the original string is untouched, and that original string is already in correct format, so the test case passes.

##### Tests performed
Tested on Mac. The error is gone. All 40 tests pass.

### DIFF
--- a/extension/src/project/readOnlyDocument.ts
+++ b/extension/src/project/readOnlyDocument.ts
@@ -87,19 +87,19 @@ export class ReadonlyDocument implements vscode.TextDocument {
         return ret;
     }
 
-    initialize(fileContent: string, languageId: string) {
-        this.fileContent = fileContent.split('\r\n').join('\n').split('\n').join('\r\n');//to make sure every line ends with \r\n
+    initialize(contentOfFile: string, langId: string) {
+        this.fileContent = contentOfFile.split('\r\n').join('\n').split('\n').join('\r\n');//to make sure every line ends with \r\n
         this.eol = vscode.EndOfLine.CRLF;
         this.eolLength = 2;
 
-        this.languageId = languageId;
+        this.languageId = langId;
 
         if (this.fileContent.length === 0) {
             this.lineCount = 0;
             this.lines = [];
         }
         else {
-            this.lines = fileContent.split('\r\n');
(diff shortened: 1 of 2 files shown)


extension/src/project/readOnlyDocument.ts
```diff
@@ -87,19 +87,19 @@ export class ReadonlyDocument implements vscode.TextDocument {
         return ret;
     }
 
-    initialize(fileContent: string, languageId: string) {
-        this.fileContent = fileContent.split('\r\n').join('\n').split('\n').join('\r\n');//to make sure every line ends with \r\n
+    initialize(rawContent: string, langId: string) {
+        this.fileContent = rawContent.split('\r\n').join('\n').split('\n').join('\r\n');//to make sure every line ends with \r\n
```
JD-Howard, line 91: I've looked at this line several times before and always kind of wondered why we weren't doing something like this:
`rawContent.replace(/[^\r]\n/g, '\r\n')`

extension/src/project/readOnlyDocument.ts
```diff
@@ -87,19 +87,19 @@ export class ReadonlyDocument implements vscode.TextDocument {
...
-        this.languageId = languageId;
+        this.languageId = langId;
 
         if (this.fileContent.length === 0) {
             this.lineCount = 0;
             this.lines = [];
         }
         else {
-            this.lines = fileContent.split('\r\n');
+            this.lines = this.fileContent.split('\r\n');
```
JD-Howard, line 102: I tested this on my mochawesome PR branch by correcting this line. It ran all the tests and opened up in the browser perfectly after implementing your OS specific fix.

extension/src/project/readOnlyDocument.ts
Sen-real: Possibly because regular expressions are normally harder to maintain. I updated the code with your suggestion.

nigma456 (approved):

PR #116, opened 2021-03-29
Add an E2E test to verify the existence of debug config items

Sen-real:
##### Objective
Add an E2E test to verify the existence of debug config items, which are:
- AutoLISP Debug: Attach
- AutoLISP Debug: Launch

They are expected to show up for user to select when starting to debug via. F5.

##### Abstractions
The following tool is used to write the test:
https://www.npmjs.com/package/vscode-extension-tester

Here's an introduction of this tool:
https://developers.redhat.com/blog/2019/11/18/new-tools-for-automating-end-to-end-tests-for-vs-code-extensions/

##### Tests performed
Ran 'npm run e2etest'. It worked as expected, i.e.:
- an instance of VS Code started
- the "Start Debugging" command was triggered via. command palette
- the existence of AutoLISP config items were verified

Locally changed the name of an expected config to make sure the test case could fail on missing that config.

NOTE:
This PR doesn't include the config to launch E2E tests from VS Code. It could be added after we figure out how to get it work.


extension/src/test/e2e/config.ts
```diff
@@ -0,0 +1,10 @@
+import { MochaOptions } from "vscode-extension-tester";
+
+const options: MochaOptions = {
+    reporter: 'spec',
+    slow: 75,
+    timeout: 10000,
+    ui: 'bdd'
```
szilvaa-adsk, line 7: Can you add some code comments please why/how you chose these? It will be useful for the maintainer.

szilvaa-adsk (commented):
It is very nice that we can now write e2e tests too.

nigma456 (approved):

extension/src/test/e2e/config.ts
Sen-real: Done. The code is updated with comments.

package.json
```diff
@@ -13,7 +13,7 @@
 	},
 	"homepage": "https://github.com/Autodesk-AutoCAD/AutoLispExt",
 	"engines": {
-		"vscode": "^1.31.1"
+		"vscode": "^1.50.0"
```
hualin-adsk, line 16: This will make the extension only support 1.50.0 or the higher version VSCODE

.vscodeignore
```diff
@@ -11,6 +11,7 @@ extension/tsconfig.json
 extension/vsc-extension-quickstart.md
 extension/node_modules/**
 test_case/**
+test-resources/**
 vendor/**
```
hualin-adsk, line 15: I see we added the package `vscode-extension-tester`, it is best to filter out from the vscode extension pacakge, to minimize the technical debt.

package.json
Sen-real: It's because of the following error:
image

Shall we apply my change or lower the version of types/vscode?

.vscodeignore
Sen-real: I didn't find it in the vsix that was made by this tester tool during the e2e test:
image

Sen-real: just checked the vsix made by pack.py too, though it's bigger, it doesn't include this new package either.

hualin-whl (approved):

szilvaa-adsk (approved):

PR #118, opened 2021-04-02
Added test cases for lisp extension format

nigma456:
##### Objective
Most complex part of the lisp extension work is formatting so we want to add more and more test cases to avoid regressions.
##### Abstractions
In this submission I have added some test cases to cover the lisp extension formatting feature.
1. The lisp extension format has 4 settings which exposed to users in the vscode setting for extension.
format.CloseParenthesisStyle : 'Same line' | 'New line with outer identation'
format.MaxLineChars || 85
format.LongListFormatStyle : 'Single Column' | 'Fill to Margin'
format.NarrowStyleIndent || 2
we tested the formatting by assigning different value for these settings.

2. To get and set values for these settings there are two ways to do.
One way is to read the configuration of these setting by following methods:
let configuration = vscode.workspace.getConfiguration('autolispext');
let formatConfigure = configuration.get('format');
then 
vscode.workspace.getConfiguration('autolispext').update('format.LongListFormatStyle','Single Column');
But it needs to either relaunch the vscode or wait several seconds until it can take effect, so this is not practical for our 
test.
The other way is to directly change the value by adding some functions exposed from the formatting code.
like 
export function resetMaxLineChars() {
    gMaxLineChars = 80;
    gHasMaxLineChars = false;
}
This is the way we used in our test cases.

3. All the 16 test cases covered various scenarios that I can image, may need to add more cases in the future.
After this submission the code coverage for format reached to 62%
image

##### Tests performed
Pass local test and CI on Github


extension/src/format/sexpression.ts
```diff
@@ -325,29 +325,75 @@ export class LispContainer extends LispAtom {
...
-      gHasSetLongListFormat = false;
-  }
-  
-  export function indentationForNarrowStyle(): number {
-      return gIndentSpaces;
-  }
+}
+
+let gMaxLineChars = 80;
+let gHasMaxLineChars = false;
```
Sen-real, line 335: This is test case only. Can we avoid doing it in feature code?

extension/src/format/sexpression.ts
```diff
@@ -325,29 +325,75 @@ export class LispContainer extends LispAtom {
...
     kWideStyleSingleCol,
     kFitToMargin
-  }
-  
-  let gMaxLineChars = 80;
-  let gIndentSpaces = 2;
-  let gClosedParenInSameLine = true;
-  let gLongListFormatAsSingleColumn = LongListFmts.kFitToMargin;
-  let gHasSetLongListFormat = false;
-  export function longListFormatAsSingleColum() {
```
Sen-real, line 339: Let's not format the entire file in a PR that contains real code changes. It takes extra time to review things that are not really changed.
If the formatting is wrong, we can submit the test case change first, then create another PR which is purely for code format.

extension/src/format/sexpression.ts
```diff
@@ -1134,31 +1180,36 @@ export class Sexpression extends LispAtom {
 
     formatting(startColumn: number, linefeed?: string): string {
 
-        gMaxLineChars = maximumLineChars();
-        if (gMaxLineChars < 60)
-            gMaxLineChars = 60;
+        if (!gHasMaxLineChars) {
```
Sen-real, line 1183: I don't know if this change is correct or not, it seems that when running this method in product, the flag is false, and then the original code that Hualin wrote won't run? is my understanding correct?
Anyway, I suggest not to change feature logic like this only for adding test cases. 
If it's really hard to test against sth. without changing the logic, we can do it via. E2E.

extension/src/format/sexpression.ts
```diff
@@ -1134,31 +1180,36 @@ export class Sexpression extends LispAtom {
...
+            if (gMaxLineChars < 60)
+                gMaxLineChars = 60;
+        }
 
-        gIndentSpaces = indentSpaces();
-        if (gIndentSpaces < 1)
-            gIndentSpaces = 1;
-        else if (gIndentSpaces > 6)
-            gIndentSpaces = 6;
+        if (!gHasIndentSpaces) {
```
Sen-real, line 1189: same as above

extension/src/format/sexpression.ts
```diff
@@ -1134,31 +1180,36 @@ export class Sexpression extends LispAtom {
...
+                gIndentSpaces = 1;
+            else if (gIndentSpaces > 6)
+                gIndentSpaces = 6;
+        }
 
-        let parenStyle = closeParenStyle();
-        if (parenStyle.toString() == "Same line")
-            gClosedParenInSameLine = true;
-        else gClosedParenInSameLine = false;
+        if (!gHasClosedParenInSameLine) {
```
Sen-real, line 1197: same as above

Sen-real (changes requested):

nigma456:
Close this PR and created a new PR
https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/121

Issue #70, opened 2021-01-26
I want to be able to configure the help link


LanDong-adsk:
**Is your feature request related to a problem? Please describe.**
when I click Open Online Help, I want to be directed to the help page with same version as my AutoCAD's.

**Describe the solution you'd like**
I want to be able to configure the online help link so that I can change as needed.

JD-Howard:
>Same version as my AutoCAD

Where do you propose this information to comes from? Are we talking:
- New user setting?
- QuickPickItem question asked once per session?
- Mandating a launch value be configured and parsing out the year?
- Using the AcadProcessFinder? (this might be windows only, but I am not sure) 

I would also point out the recent change that imbedded the year at the time the JSON data file was generated, did add a large degree of bullet proofing around Autodesk's web team changing the GUID's on us. This enhancement would mitigate that security blanket.

hualin-adsk:
After we discussed with testers and user experiences assistant, we thought that the extension can be used in old AutoCAD like AutoCAD2021, user MAY want to access the corresponding help page instead of the current year or hard code year.
So we created this issue here, and plan to hear feedback from users after released this version extension.

hansWurst-creator:
I would like it to be automatically configured if the setting "autolispext.debug.LaunchProgram" is set.
A once per session question gets annoying very fast.
A user setting would probably be the easiest to implement and should work fine for now.

Sen-real:
Resolved in AutoLISP extension v1.5.0

PR #95, opened 2021-03-13
Lisp Container & Lisp Formatting Separation

JD-Howard:
##### Objectives
* Create a separation between the Formatting & Structure contexts
* Create an interface that can be used to generically represent Lisp structures
* Increase the Test coverage of the Structural types

##### Abstractions
The added `ILispFragment` interface has an optional `LispContainer.body` pass through property. This lets us make an attempt at using `LispContainer` functionality without ever doing any type checking. It would go something like this `item.body?.getExpressionFromPos(pos)` and we'll either have a result or we won't. From that point you aren't really checking the type of value returned, but rather if it has a value, then you can do more stuff.

##### Tests performed
There are several extensive test files in there. Some of which were previously named other things that now made less sense given the changes in this PR. This created a lot more code coverage for anything that is meaningfully associated with this PR. I also manually tested the context menu formatting options and F12 extensively. The testing of F12 identified a few bugs in the `LispParser.getDocumentExpression()` and has been fixed in this PR. 

**DocumentExpression.test.ts** -> renamed to **DocumentContainer.test.ts** and has minor changes
**Sexpression.test.ts** was renamed to **LispContainer.test.ts** has extensive changes
**LispAtom.test.ts** is brand new

##### Screen shot
Not applicable

##### Other Stuff
You will noticed that the `isValidSexpr()` and `format()` functions have been removed from LispAtom. I made every attempt not to disturb the `Sexpression` formatting code, but had to for at least removing these from the LispAtom. Other than those, I really only moved "newer" features that have nothing to do with the formatting behaviors out of the Sexpression.

Issue #90, opened 2021-02-24
different auto format between foreach and vlax-for


hansWurst-creator:
Currently the autoformating of foreach and vlax-for is creating the following result:
![image](https://user-images.githubusercontent.com/70693364/108996409-9f08da80-769e-11eb-90d9-a0667af5c15e.png)
Something that is bothering me, is that the foreach and vlax-for look differently with auto format. With foreach first the iteration symbol and then the list through which is iterated are all put in one line. This is also the same syntax you see in many other languages (eg. vba, c# etc.). With vlax-for the collection is put on a separate line.
This makes it a bit harder to understand what is going on in my opinion.

I would prefer if autoformating would have this result:
![image](https://user-images.githubusercontent.com/70693364/108997635-1f7c0b00-76a0-11eb-9ddc-4b57b6d17ded.png)

Sen-real:
@hansWurst-creator :
do you mean to provide auto-formatting like this?
![image](https://user-images.githubusercontent.com/57521938/109292890-03c26180-7866-11eb-9d3f-a4001504f230.png)

hansWurst-creator:
yeah sorry that is what I meant. Hmm I must have copied the wrong picture.

JD-Howard:
I can fix this when Parsing & Formatting get abstracted away from each other. I tested changing:
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/2d7759d789b8a3239fc32c487b0746112163b5da/extension/src/format/sexpression.ts#L844
into
`else if (opName == "foreach" || opName == "vlax-for") {`
seems to work fine

hansWurst-creator:
this is fixed in Version 1.50

PR #119, opened 2021-04-03
Resolved Issue #70

JD-Howard:
#### Objective
Issue #70 requested user control over the help URL year. The proposed solution was to extract this from the Launch Path Configuration. This implements that request and a few other minor refactors.

#### Abstractions
The function we had for reading our Configuration Settings was not exported and lived in the `debug.ts` file. This was relocated to `resources.ts`, flagged as an export and the `debug.ts` was updated to use the new shared location.

#### Tests performed
Manually tested this and it pulled up the expected result by changing the launch configurations to:
* empty
* C:\Program Files\Autodesk\AutoCAD 2019\acad.exe
* C:\Program Files\Autodesk\AutoCAD 2020\acad.exe
* C:\Program Files\Autodesk\AutoCAD 2021\acad.exe
* C:\Program Files\Autodesk\AutoCAD 2022\acad.exe
* C:\Program Files\Autodesk\AutoCAD 2023\acad.exe (this produces not found)

### DIFF
--- a/extension/src/context.ts
+++ b/extension/src/context.ts
@@ -29,7 +29,7 @@ export class ContextManager{
 	get ExtPath(): string { 
 		return this._ctx ? this._ctx.extensionPath : ""; 
 	}
-	get Data() { 
+	get Resources() { 
 		return Resources; 
 	}
 
--- a/extension/src/format/sexpression.ts
+++ b/extension/src/format/sexpression.ts
@@ -247,11 +247,6 @@ export class LispContainer extends LispAtom {
     }
 
 
-    // Returns true if this LispContainer encapsulates the provided Position
-    contains(position: Position): boolean {
-        return this.getRange().contains(position);
-    }
-
 
     // This version was necessary to properly alternate over SETQ Name vs Value 
     private isValidForSetq(atom: ILispFragment): boolean {
--- a/extension/src/format/sexpression.ts
+++ b/extension/src/format/sexpression.ts
@@ -1100,7 +1100,7 @@ export class Sexpression extends LispAtom {
                     else if (opName == "setq") {
                         return exp.formatSetq(startColumn);
                     }
-                    else if (opName == "foreach") {
+                    else if (opName === "foreach" || opName === "vlax-for") {
                         return exp.formatForeach(startColumn);
                     }
                     else if (opName == "defun" || opName == "defun-q") {
--- a/extension/src/help/openWebHelp.ts
+++ b/extension/src/help/openWebHelp.ts
@@ -1,5 +1,5 @@
 import * as vscode from 'vscode';
-import { AutoLispExt } from '../extension';
+import { getExtensionSettingString } from '../resources';
 import { IJsonLoadable, webHelpContainer } from "../resources";
 
 
@@ -37,7 +37,7 @@ export class WebHelpLibrary implements IJsonLoadable {
 	// Issue #70 requested user configuration of the help year
 	// 			 this implements the proposed solution of extracting it from the launch configuration
 	private getYearFromSettings(): string|null {
-		let acad = AutoLispExt.Resources.getExtensionSettingString('debug.LaunchProgram');
+		let acad = getExtensionSettingString('debug.LaunchProgram');
 		let qualifier = /(?<=AUTOCAD\s*)20\d{2}(?!ACAD.EXE)/i;
 		if (qualifier.test(acad)) {
 			return qualifier.exec(acad)[0];
--- a/extension/src/help/openWebHelp.ts
+++ b/extension/src/help/openWebHelp.ts
@@ -27,28 +27,23 @@ export class WebHelpLibrary implements IJsonLoadable {
 	functions: Dictionary<WebHelpFunction> = {};
 	ambiguousFunctions: Dictionary<WebHelpFunction[]> = {};
 	enumerators: Dictionary<string> = {};	
-	_fallbackYear: string;
+	_jsonYear: string = '';
+	
 	get year(): string {
-		// This was converted to a getter because WebHelpLibrary only loads once on start
-		// It would of required restart after changing the launch configuration to update the year
-		return this.getYearFromSettings() ?? this._fallbackYear;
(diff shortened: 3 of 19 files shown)


extension/src/context.ts
```diff
@@ -29,7 +29,7 @@ export class ContextManager{
 	get ExtPath(): string { 
 		return this._ctx ? this._ctx.extensionPath : ""; 
 	}
-	get Data() { 
+	get Resources() {
```
JD-Howard, line 32: Note, nothing in the Repo was using this, it was exposed for planned changes to AutoCompletion

extension/src/format/sexpression.ts
```diff
@@ -247,11 +247,6 @@ export class LispContainer extends LispAtom {
     }
 
 
-    // Returns true if this LispContainer encapsulates the provided Position
-    contains(position: Position): boolean {
-        return this.getRange().contains(position);
-    }
-
```
JD-Howard, line 254: This was debt discussed in PR #95

JD-Howard:
Noticed the CI failed, figured out I had created a race condition accessing something before it was actually loaded. Simple fix that actually resulted in less changes to the `debug.ts` so it was win-win. Should be fine now.

Sen-real:
@JD-Howard this is cool! please let me know when this PR is ready to merge.

Sen-real:
The code change looks good to me, but I'm not sure whether Josh is ready to let us merge.

extension/src/help/openWebHelp.ts
```diff
@@ -26,11 +27,28 @@ export class WebHelpLibrary implements IJsonLoadable {
...
+	get year(): string {
+		// This was converted to a getter because WebHelpLibrary only loads once on start
+		// It would of required restart after changing the launch configuration to update the year
+		return this.getYearFromSettings() ?? this._fallbackYear;
+	}
+
+	// Issue #70 requested user configuration of the help year
+	//           this implements the proposed solution of extracting it from the launch configuration
+	private getYearFromSettings(): string|null {
+		let acad = getExtensionSettingString('debug.LaunchProgram');
```
hualin-whl, line 40: Maybe we need to add an independent setting for help? it looks like this break the [SRP ](https://en.wikipedia.org/wiki/Single-responsibility_principle)rule.

JD-Howard: Hi Hualin,
I agree this smells a bit, but wouldn't having a dedicated setting vs extracting it from an existing setting be pretty much the same operation? I guess I don't know how to fix the SRP problem since dealing with any setting is going to inject a dependency.

JD-Howard:
I think this is generally good to go, but I don't know how to resolve Hualin's SRP comment, if anyone has some suggestions for that please share them, if you also don't see a way to handle it, then you can merge the PR.

JD-Howard:
Apparently I shouldn't have identified some more PR#95 debt and fixed it. Super insignificant change, but its not passing CI now. Looks like nothing has passed CI since #120 _To hook e2e tests on CI_

extension/src/format/sexpression.ts
```diff
@@ -1105,7 +1100,7 @@ export class Sexpression extends LispAtom {
                     else if (opName == "setq") {
                         return exp.formatSetq(startColumn);
                     }
-                    else if (opName == "foreach") {
+                    else if (opName === "foreach" || opName === "vlax-for") {
```
JD-Howard, line 1103: This was debt discussed in Issue #90 that I nearly forgot about.

extension/src/help/openWebHelp.ts
```diff
@@ -26,11 +27,28 @@ export class WebHelpLibrary implements IJsonLoadable {
...
+		let qualifier = /(?<=AUTOCAD\s*)20\d{2}(?!ACAD.EXE)/i;
```
JD-Howard, line 41: Umm, just realized this probably doesn't work for Mac. Now that I have one, I just downloaded the 2022 dmg today, guess I should go do something with it... Maybe this is why Hualin was leaning towards a new setting property instead of a using the launch config?

JD-Howard:
Sorry to make this a cluster, but Hualin was right, we needed a new setting to handle this. You now have a slew of i18 empty placeholder changes in this PR. I did that just to keep the files consistently ordered/make sure the localization team can easily see what needs to be populated.

extension/src/help/openWebHelp.ts
Sen-real: I asked Mac developers. The path on Mac is sth. like this:
/Applications/Autodesk/AutoCAD 2022/AutoCAD 2022.app

Sen-real: Though we could check the OS in code like this:
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/a4b83161d97835111811c0d768431ee6a03627d1/extension/src/platform.ts#L16

But I think Hualin's suggestion is to define a new setting for OpenWebHelp, so that user can manually define what ever version that is proper to use. 

And it's also possible that a user will never use launch mode - e.g. I always use the Attach mode. In this case, the launch setting could be wrong or empty.

Sen-real:
> Apparently I shouldn't have identified some more PR#95 debt and fixed it. Super insignificant change, but its not passing CI now. Looks like nothing has passed CI since #120 _To hook e2e tests on CI_ was merged.

I reran the job and it's green now.
It was not a CI failure, instead I think the CI died on your job...

Sen-real (approved):

extension/src/help/openWebHelp.ts
hualin-whl: If there is a third time we want to re-use this setting, then that is the time to split this setting to each individual settings.

extension/src/format/sexpression.ts
hualin-whl, line 1103: I am thinking that maybe this is naive fix here. In my memory I remember that in OLD vlide there is a setting for if the first word is longer than 9 chars, it should use NARROW formating style, otherwise it should use the WIDE formatting style.

PR #121, opened 2021-04-06
Added test cases for lisp extension format with mock method

nigma456:
##### Objective
Most complex part of the lisp extension work is formatting so we want to add more and more test cases to avoid regressions.

##### Abstractions
In this submission I have added some test cases to cover the lisp extension formatting feature.

1. The lisp extension format has 4 settings which exposed to users in the vscode setting for extension.
format.CloseParenthesisStyle : 'Same line' | 'New line with outer identation'
format.MaxLineChars || 85
format.LongListFormatStyle : 'Single Column' | 'Fill to Margin'
format.NarrowStyleIndent || 2
we tested the formatting by assigning different value for these settings.

2. Since in the Github CI virtual enviornment the lisp extension can't be activated when running the unit test we have to use mock method to do the test.

3. All the 16 test cases covered various scenarios that I can image, may need to add more cases in the future.
After this submission the code coverage for format reached to 62%
image
image

4. Next will try to add test cases for format selection.

##### Tests performed
Pass local test and CI on Github


package.json
```diff
@@ -615,6 +615,7 @@
 	"scripts": {
 		"compile": "tsc -p ./",
 		"watch": "tsc -p -watch ./",
+		"pretest": "winCompile.bat",
```
Sen-real, line 618: this is windows only; for mac it's ./macCompile.sh
I don't know what's the best way to make it cross platform, but I found the same question here:
https://stackoverflow.com/questions/45082648/npm-package-json-os-specific-script

extension/src/format/sexpression.ts
```diff
@@ -1145,13 +1145,13 @@ export class Sexpression extends LispAtom {
             gIndentSpaces = 6;
 
         let parenStyle = closeParenStyle();
-        if (parenStyle.toString() == "Same line")
+        if (parenStyle.toString().toLocaleLowerCase() == "same line")
```
Sen-real, line 1148: Is there a special reason for toLocaleLowerCase() ? according to the old code, the left operator should be an English string.

nigma456: This is a mistake, it should be .tolowercase()

package.json
nigma456: Using python or JS to do this job.

LanDong-adsk: just corious, winCompile.bat looks not used anywhere. what is this file used for?

LanDong-adsk: after reading the code briefly, I don't see any need for the unit test to call this command. Anything I was missing?

LanDong-adsk:
Are these lisp files OK to be published to public? any intellectual property concerns?

package.json
JD-Howard: If you want to run unit tests on any of the Help related functions, you'll need to make sure the `webHelpAbstraction.json` file gets copied where it belongs somehow, but other than that the `acadProcessFinder.exe` copy can't be of any use in CI.

I had suggested we set these copy operations up using a python script hooked into the tasks.json a while back, but I am not sure if that will play well with your plans. I just know its worked pretty well for me out of CI.
https://github.com/JD-Howard/FabCOD/blob/main/.vscode/tasks.json

Sen-real: @LanDong-adsk winCompile.bat (and ./macCompile.sh) are used for developers to do a clean build, i.e. to clean up the out folder first, and then npm run compile.
Long ago we got some weird problem on dev machine, which was gone after locally doing a clean build. Then we added these two scripts.
I once wanted to explain the use of these scripts in DevReadme, but Hualin stopped me because he thought that the new comers could be confused by too much information.

nigma456: winCompile.bat is necessary to run test especially during the test cases writing since the changed and compiled script will impact the result and I have seen some problems related to this recently.
But for Github CI enviornment since the env is clean it is unnecessary to run it.

.github/codeql/codeql-config.yml
```diff
@@ -1,10 +1,12 @@
...
+  - uses: security-and-quality
 
 paths:
-- '**/extension/src/**'
+  - '**/extension/src/**'
 
 paths-ignore:
-- '**/.vscode-test/**'
+  - '**/.vscode-test/**'
+  - '**/AutoLispExt/out/**/*.js'
```
Sen-real, line 11: this is already in another PR

extension/src/extension.ts
```diff
@@ -30,6 +30,7 @@ loadAllResources();
 
 export function activate(context: vscode.ExtensionContext) {
 	AutoLispExt.initialize(context);	
+	console.log(`--------------extension.ts----function activate--------------`);
```
Sen-real, line 33: can we revert this log?

package.json
```diff
@@ -639,9 +639,11 @@
 		"nyc": "^15.1.0",
 		"run-sequence": "^2.2.1",
 		"source-map-resolve": "0.6.0",
+		"ts-mock-imports": "^1.3.3",
```
Sen-real, line 642: I think we need to update package-lock.json too

extension/src/test/suite/Formatter_mock.test.ts
```diff
@@ -0,0 +1,392 @@
...
+    if (closeParenStyleStub) {
+      closeParenStyleStub.restore();
+    }
+    closeParenStyleStub = ImportMock.mockFunction(
+      fmtConfig,
+      "closeParenStyle",
+      closeParenStyle
+    );
+  }
+  async function setMaxLineChars(maxchar: number) {
```
Sen-real, line 372: there's no 'await' in this method, so I there's no need to mark it as async or to use await to call this method

extension/src/extension.ts
nigma456: Yes, right.

package.json
nigma456: Why? The package-lock.json does not need to upload, right?

Sen-real: Please check the history of package-lock.json, we've been updating this file all the time.

.github/codeql/codeql-config.yml
nigma456: Yes, I need to merge that PR otherwise it will take very long time to run the codeql.

extension/src/test/suite/Formatter_mock.test.ts
LanDong-adsk: I second that.

nigma456: Will update.

extension/src/test/suite/Formatter_mock.test.ts
```diff
@@ -0,0 +1,392 @@
...
+  });
+
+  suiteTeardown(async () => {
+    closeParenStyleStub.restore();
+    maximumLineCharsStub.restore();
+    longListFormatStyleStub.restore();
+    indentSpacesStub.restore();
+  });
+  suiteSetup(async () => {
+    closeParenStyleStub = ImportMock.mockFunction(
```
LanDong-adsk, line 93: you can just call setClosedParenInSameLine("...") here.
Same as other places. 
We need to minimize duplicated code.

nigma456: Good point.

Sen-real (approved):

Sen-real:
@nigma456 it looks good, but please merge with latest main code first

PR #122, opened 2021-04-08
Add test cases for lisp extension Autocompletion function

nigma456:
##### Objective
Add test cases for lisp extension Autocompletion function

##### Abstractions
Added test cases for lisp extension Autocompletion function:
getLispAndDclCompletions()
getCmdAndVarsCompletionCandidates()
Does not cover the function:
getMatchingWord()


JD-Howard:
AutoCompletion is something on my agenda to enhance in the near future. If there is anything in there that needs to be broke apart for easier/more complete testing, let me know.

General question about testing:
_Should we be making some kind of standardized singular pass-through to expose a file or classes non-exported/private functions/variable to testing?_ I've been wondering for a while, how I could obtain good test coverage while maintaining the intended "shared" usefulness context. What I am suggesting here could be far worse than what you have and I am just spitballing a concept. In reference to this PR, it would probably look like this:
```
export class MochaTest_completion {
    static get getCmdAndVarsCompletionCandidates() { 
        return getCmdAndVarsCompletionCandidates; 
    }
    static get getCompletionCandidates() { 
        return getCompletionCandidates; 
    }
    static get getMatchingWord() { 
        return getMatchingWord; 
    }
    static get getLispAndDclCompletions () { 
        return getLispAndDclCompletions; 
    }
}
```

On the test file it would get used like this:
```
import { MochaTest_completion } from "./completion/autocompletionProvider";
MochaTest_completion .getLispAndDclCompletions(whatever);
```
IDK if this is diplorable duplication or not, but something does feel wrong about arbitrarily setting everything to a fully exposed status exclusively to make it testable.

nigma456:
@JD-Howard Yes, we may need to either refactor the source code or expose wrapped method to facilitate the test better.

PR #124, opened 2021-04-09
Fix the CodeQL very slow issue

nigma456:
##### Objective
Recently it takes very long time(~50 minutes) to run the codeql on Github CI due to the test-resources folder
##### Abstractions
Add the test-resources folder to the configure file codeql-config.yml so that the codeql will not scan it.
##### Tests performed
Passed local test and Github CI

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -7,8 +7,7 @@ paths:
 - '**/extension/src/**'
 
 paths-ignore:
-- '**/.vscode-test/**'
+  - '**/.vscode-test/**'
   - '**/.vscode-test/**'
   - '**/AutoLispExt/out/**/*.js'
   - '**/test-resources/**'
-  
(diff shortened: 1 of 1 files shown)


.github/codeql/codeql-config.yml
```diff
@@ -7,4 +7,7 @@ paths:
 - '**/extension/src/**'
 
 paths-ignore:
-- '**/.vscode-test/**'
+  - '**/.vscode-test/**'
+  - '**/.vscode-test/**'
```
Sen-real, line 11: this is duplicate

.github/codeql/codeql-config.yml
```diff
@@ -7,4 +7,7 @@ paths:
...
+  - '**/AutoLispExt/out/**/*.js'
```
Sen-real, line 12: I guess you meant **/AutolispExt/out/test/**

nigma456: No, we should not scan the compiled js file under out folder.

Sen-real (approved):

.github/codeql/codeql-config.yml
Sen-real: I see.

PR #127, opened 2021-04-12
Add test cases for autocompletion

nigma456:
##### Objective
Add test cases for lisp extension Autocompletion feature
##### Abstractions
1. Change the fs.readFile to fs.readFileSync so that the variables [internalLispFuncs,internalDclKeys etc] can got value before the mocha test.
This may impact a little bit performance of launch extension for end client users but this delay will be very tiny.

2. Cover getCmdAndVarsCompletionCandidates() and getLispAndDclCompletions()
3. Does not cover getMatchingWord() since it is related to vscode editor and it will return error when running into the method document.getWordRangeAtPosition(position) after I created and opened a lisp file to test.
4. Need to run on Mac os to verify since there are difference between Mac and Win32 OS.
5. After the submission the code coverage for autocompletion will be 53%

image

##### Tests performed
Passed Local and Github CI


extension/src/completion/autocompletionProvider.ts
```diff
@@ -114,9 +114,14 @@ function getMatchingWord(document: vscode.TextDocument, position: vscode.Positio
...
 }
 
-function getLispAndDclCompletions(document: vscode.TextDocument, word: string, isupper: boolean): vscode.CompletionItem[] {
+export function getLispAndDclCompletions(document: vscode.TextDocument, word: string, isupper: boolean): vscode.CompletionItem[] {
     let currentLSPDoc = document.fileName;
     let ext = currentLSPDoc.substring(currentLSPDoc.length - 4, currentLSPDoc.length).toUpperCase();
+    if(internalLispFuncs == [] || internalDclKeys == [] || allCmdsAndSysvars == [] || winOnlyListFuncPrefix == []){
+        console.log('Reload the resource since the value are empty');
+        
+        loadAllResources();
```
JD-Howard, line 123: `loadAllResources()` is supposed to happen once on extension load, maybe it needs to stay where it is (in the extension.ts) and just be turned into an awaited function? IE, you may have a unfulfilled promise issue.

I just know it can't be here, this function gets called constantly because it is activated by user typing.

nigma456: The problem here is that on Github CI virtual enviornment I found the extension is not actually activated so I tried to load it before using it.

JD-Howard: @nigma456 , I didn't see your comment and edited mine. So, making a reply.
looking at this further, `resources.ts` has a lot of FS ReadFile() going on that may need to be turned into [ReadFileSync](https://nodejs.dev/learn/reading-files-with-nodejs).

PR #128, opened 2021-04-12
Added test cases for lisp extension autocompletion

nigma456:
##### Objective
Add test cases for lisp extension Autocompletion feature

##### Abstractions
1. Change the fs.readFile to fs.readFileSync so that the variables [internalLispFuncs,internalDclKeys etc] can got value before the mocha test.
This may impact a little bit performance of launch extension for end client users but this delay will be very tiny.

2. Cover getCmdAndVarsCompletionCandidates() and getLispAndDclCompletions()
3. Does not cover getMatchingWord() since it is related to vscode editor and it will return error when running into the method document.getWordRangeAtPosition(position) after I created and opened a lisp file to test.
4. Need to run on Mac os to verify since there are difference between Mac and Win32 OS.
5. After the submission the code coverage for autocompletion will be 53%
image

##### Tests performed
Passed Local and Github CI


extension/src/resources.ts
```diff
@@ -40,28 +38,27 @@ function readJsonDataFile(datafile: string, intoObject: IJsonLoadable): void {
...
 	var dataPath = path.resolve(__dirname, datafile);
-	fs.readFile(dataPath, "utf8", function(err: Error, data: string) {
-		if (err === null) {
-			if (data.includes("\r\n")) {
-				action(data.split("\r\n"));
-			}
-			else {
-				action(data.split("\n"));
-			}
+	let data = fs.readFileSync(dataPath,{encoding:'utf8', flag:'r'});
```
Sen-real, line 43: This should be placed in the try block. It will throw exception on failure.

extension/src/resources.ts
```diff
@@ -40,28 +38,27 @@ function readJsonDataFile(datafile: string, intoObject: IJsonLoadable): void {
...
+		console.log(error);
+	}
 }
 
 
 function readDataFileByDelimiter(datafile: string, delimiter: string, action: (item: string) => void) {
-	var fs = require("fs");
 	var dataPath = path.resolve(__dirname, datafile);
-	fs.readFile(dataPath, "utf8", function(err: Error, data: string) {
+	let data = fs.readFileSync(dataPath,{encoding:'utf8', flag:'r'});
```
Sen-real, line 59: Same as above.

extension/src/test/suite/AutoCompletion.test.ts
```diff
@@ -0,0 +1,273 @@
...
+  let suggestLabel = [];
+  suggestionList.forEach((item) => {
+    suggestLabel.push(item.label);
+  });
+  return suggestLabel;
+}
+
+suite("AutoCompletion Tests", function () {
+  // Windows only functions (vla-,vlax-,vlr-,vl-load-com,vl-load-reactors,vlisp-)
+  before(() => {
```
Sen-real, line 65: it's "setup" for suite, and "before" for "describe"

Sen-real (changes requested):

extension/src/test/suite/AutoCompletion.test.ts
Sen-real: https://mochajs.org/#configuring-mocha-nodejs
image
image

nigma456: Ok, I will update.

extension/src/bracketMatch/bracketMatching.ts
```diff
@@ -0,0 +1,368 @@
+import * as vscode from 'vscode';
+
```
Sen-real, line 2: is this file added by accident?

extension/src/resources.ts
```diff
@@ -40,28 +38,27 @@ function readJsonDataFile(datafile: string, intoObject: IJsonLoadable): void {
...
 }
 
 
 function readDataFileByDelimiter(datafile: string, delimiter: string, action: (item: string) => void) {
-	var fs = require("fs");
 	var dataPath = path.resolve(__dirname, datafile);
-	fs.readFile(dataPath, "utf8", function(err: Error, data: string) {
+	try {
+		var data = fs.readFileSync(dataPath,{encoding:'utf8', flag:'r'});
 		var lineList = new Array<String>();
```
Sen-real, line 61: I think the code scanning is correct, there's no need to initialize lineList

extension/src/bracketMatch/bracketMatching.ts
nigma456: Yes, it is. Looks like the git source control in vscode is doing the git add cross branches.

extension/src/resources.ts
nigma456: yes, I have changed it to var lineList : Array<String>

Sen-real (approved):

PR #130, opened 2021-04-13
Add the format selection unit test cases

nigma456:
##### Objective
Added the format selection unit test cases.
##### Abstractions
Form a vscode position to get the  selection and then do the format.
##### Tests performed
Passed local test and Github CI

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,7 +27,11 @@
             ],
             "outFiles": [
                 "${workspaceFolder}/out/test/**/*.js"
-            ]
+            ],
+            "skipFiles": [
+                "${workspaceFolder}/node_modules/**/*.js",
+                "<node_internals>/**/*.js"
+              ]
         },
         {
             "name": "Extension Tests + Reporting",
--- a/extension/src/test/suite/Formatter.test.ts
+++ b/extension/src/test/suite/Formatter.test.ts
@@ -76,7 +76,7 @@ suite("Lisp Formatter mock Tests", function () {
 
   setup(()=>{
     resetDefault();
-  })
+  });
 
   teardown(() => {
     internalLispFuncsStub.restore();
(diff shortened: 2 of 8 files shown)


extension/src/test/suite/Formatter.test.ts
```diff
@@ -225,6 +224,7 @@ suite("Lisp Formatter mock Tests", function () {
     // Test long list and big file - chinaMap.lsp
     // This test will take long time
     let i = 10;
+    this.timeout(20000);
```
LanDong-adsk, line 227: timeout?

nigma456: Since this file is a little big we should have more time for this single case to run the format.

.vscode/launch.json
```diff
@@ -27,7 +27,11 @@
             ],
             "outFiles": [
                 "${workspaceFolder}/out/test/**/*.js"
-            ]
+            ],
+            "skipFiles": [
+                "${workspaceFolder}/node_modules/**/*.js",
```
Sen-real, line 32: why are we adding this?

nigma456: Just don't want to debug into so many external files during unit test debug.

extension/src/test/suite/Formatter.test.ts
```diff
@@ -62,7 +62,7 @@ suite("Lisp Formatter mock Tests", function () {
   let internalLispFuncsStub;
   let internalOperators;
 
-  setup(() => {
+  setup(async () => {
```
Sen-real, line 65: I think we should change "setup" to "suiteSetup" (like 'before' in BDD), and change "beforeEach" to "setup".
The code from line 65 to 75 should be done only once.

nigma456: No, the restore should be run for each case otherwise it will fail.

Sen-real: Check the code. You're merging two methods into one, but I don't think all the code are needed to run before each test.

nigma456: Will update with the correct suitesetup.

Sen-real (approved):

PR #131, opened 2021-04-13
Add autoindent unit test cases

nigma456:
##### Objective
Add autoindent unit test cases
##### Abstractions
The autoIndent.ts is used for user input in vscode editor to decide the auto indentation when press enter under all kinds of cases.
The autoindentation logic is complex. There are lots of minor cases need to be considerred.
After this PR the code coverage for autoindent.ts achieved to 76%

image

##### Tests performed
Passed Local and Github tests


extension/src/test/suite/AutoIndent.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+}
+
+suite("Autoindent Tests", function () {
+  // Autoindent is used for input in the vscode editor
+  suiteSetup(() => {
+    createFakeTextDcoument();
+  });
+
+  test("Autoindent for defun enter should have 2 indentation", async function () {
```
LanDong-adsk, line 45: I don't know. 
From the lisp file above, defun has 0 indent and setq has 2 indent, but the test says both has 2. Is this expected?

nigma456: Yes, for both defun and setq under this context content it should have 2 indentation when press enter:
(defun c:indentTest (/ a b) 
 |(setq a 3)
 |

extension/src/test/suite/helper.ts
```diff
@@ -0,0 +1,12 @@
+
```
Sen-real, line 1: is this file used anywhere?

extension/src/test/suite/AutoIndent.test.ts
```diff
@@ -0,0 +1,292 @@
...
+let indentTestFile = "";
+let indentTestDoc: vscode.TextDocument;
+
+fs.mkdir(outputDir, { recursive: true }, (err) => {
+  if (err) {
+    return console.error(err);
+  }
+});
+
+function createFakeTextDcoument() {
```
Sen-real, line 19: I suggest "openLspTestFile"

extension/src/test/suite/AutoIndent.test.ts
```diff
@@ -0,0 +1,292 @@
...
+  indentTestFile = path.join(testDir + "/SourceFile/autoIndentTestFile.lsp");
+  try {
+    if (fs.existsSync(indentTestFile)) {
+      indentTestDoc = ReadonlyDocument.open(indentTestFile);
+    }
```
Sen-real, line 24: it should fail on else

extension/src/test/suite/AutoIndent.test.ts
```diff
@@ -0,0 +1,292 @@
...
+      let cursorPos2d = new vscode.Position(0, 28);
+      let lineIndentSpace = getIndent(cursorPos2d);
+      assert.isTrue(lineIndentSpace == 2);
+    } catch (err) {
+      assert.fail(`Autoindent test for ${fn} failed`);
+    }
+  });
+
+  test("Autoindent for setq enter should have 2 indentation", async function () {
+    let fn = "setq enter";
```
Sen-real, line 60: why can't we just use plain text in the call to assert.fail?

extension/src/test/suite/helper.ts
nigma456: Yes, I am trying to move more and more public functions used for unit test here.
I am using this module during the test case writting.

extension/src/test/suite/AutoIndent.test.ts
nigma456: Yes, But I have put this file there and don't allow it fail. but anyway I can add some warning message here.

extension/src/test/suite/AutoIndent.test.ts
nigma456: Do you mean to open a .txt file ?

extension/src/test/suite/AutoIndent.test.ts
nigma456: Yes, renamed.

extension/src/test/suite/helper.ts
Sen-real: I don't see this file imported in other file, and I don't see "sleep" or "princ" out of this file. Is there code on your machine that is not yet uploaded?

extension/src/test/suite/AutoIndent.test.ts
Sen-real: I mean instead of "assert.fail(`Autoindent test for ${fn} failed`);"
you could just use "assert.fail(`Autoindent test for setq enter failed`);"

nigma456: This will reduce the work of hard copy the message. I just need to copy this sentense without modification for every test cases.

extension/src/test/suite/helper.ts
nigma456: As I said during my test cases writting I have imported and used this file to quick log messages I needed.
But once git push I removed these usages.

extension/src/test/suite/AutoIndent.test.ts
Sen-real: No it will not reduce the work, as you then will have to change this code:
 let fn = "setq enter";

nigma456: This is a keyword for this test case so I want to put this here in front of the test.

Sen-real (approved):

PR #136, opened 2021-04-15
Add test cases for getMatchingWord()

nigma456:
##### Objective
Added test cases for getMatchingWord() in the autocompletion.test.ts
##### Abstractions
This function is used by the autocompletion feature and it is not covered before since we wrongly used the readOnlyDocument class which is defined only for project feature use. 
There are some limitation for this readOnlyDocument class, like it does not implement the getWordRangeAtPosition() method which prevent writing the getMatchingWord test cases before.

In this submission I simply implement the function readFile2TextDocument() which will read a file to a vscode.TextDocument object so that we can use it later.
In most of our unit test cases we should  use this readFile2TextDocument() methd instead of readOnlyDocument.

2. Delete used folder /test_cases


extension/src/test/suite/helper.ts
```diff
@@ -10,3 +11,12 @@ export function princ(str : any){
     var value = str[name];
 	console.log(`-----the ${name} is ${value.toString()}-----`);
 }
+//This method is to read a file and return the vscode.TextDocument object 
+export async function readFile2TextDocument(filepath: string) : Promise<vscode.TextDocument | undefined> {
+	let doc : vscode.TextDocument;
```
Sen-real, line 16: this variable should be removed as the "Code scanning" suggests

extension/src/test/suite/helper.ts
```diff
@@ -10,3 +11,12 @@ export function princ(str : any){
...
+	try {
+        return doc = await vscode.workspace.openTextDocument(filepath);
+      } catch (error) {
+        return undefined;
```
Sen-real, line 20: how about console.log(error) before return?

Sen-real (approved):

extension/src/test/suite/helper.ts
nigma456: Yes, I will add this.

extension/src/test/suite/helper.ts
nigma456: Will do a minor change to keep this variable.

Sen-real (approved):

Issue #129, opened 2021-04-12
Question: AutoLispDebugAdapter not found


WildOrangutan:
Hey, I'm getting error below. I can't find any information online, how to fix it. My autocad path and process are set correctly.

I'm using version Autocad 2020. How do I know which versions supports this adapter?

> AutoLispDebugAdapter.exe doesn’t exist. Verify that the file exists in the same folder as that for the product specified in the launch.json file.

nigma456:
We support the vscode extension debugger since AutoCAD 2021. You can check the sysvar lispsys. If it is available then it should support the vscode debugger.

LanDong-adsk:
@Sen-real , can we update the error message to indicate it only works after 2021?

Sen-real:
@LanDong-adsk 
good idea

Sen-real:
Close this issue. Solved in PR129.

vargekt:
well, 2025 and still no update in that error message to show 2021 or higher requirement, nice work guys/girls.....and what's funny, from 2021 u can't permanently buy acad, just a subscription, what a coincidence...

PR #137, opened 2021-04-15
Update an error message to include the requirement on ACAD version

Sen-real:
##### Objective
This is to solve this issue: https://github.com/Autodesk-AutoCAD/AutoLispExt/issues/129

##### Abstractions
This is to solve this issue: https://github.com/Autodesk-AutoCAD/AutoLispExt/issues/129
Currently when starting to debug LISP with an old version of AutoCAD, it will fail on file-not-found-error of debug adapter. The error message doesn't tell user that it's possibly because that a newer version of AutoCAD is needed.

A short hint is added into the error message for this possibility.

##### Tests performed
The bug is gone.

### DIFF
--- a/i18n/enu/out/debug.i18n.json
+++ b/i18n/enu/out/debug.i18n.json
@@ -1,5 +1,5 @@
 {
-  "autolispext.debug.nodap": "doesn’t exist. Verify that the file exists in the same folder as that for the product specified in the launch.json file.",
+  "autolispext.debug.nodap": "doesn’t exist. Use a release later than AutoCAD 2020. Verify that the file exists in the same folder as that for the product selected to attach.",
   "autolispext.debug.noacad": "doesn’t exist. Verify and correct the folder path to the product executable.",
   "autolispext.debug.acad.nosupport": "This instance of AutoCAD doesn’t support debugging AutoLISP files, use a release later than AutoCAD 2020.",
   "autolispext.debug.launchjson.path": "Specify the absolute path to the product with the Path attribute of the launch.json file.",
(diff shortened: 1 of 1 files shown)


i18n/enu/out/debug.i18n.json
```diff
@@ -1,5 +1,5 @@
 {
-  "autolispext.debug.nodap": "doesn’t exist. Verify that the file exists in the same folder as that for the product specified in the launch.json file.",
+  "autolispext.debug.nodap": "doesn’t exist. Use a release later than AutoCAD 2020. Verify that the file exists in the same folder with the product to debug.",
```
nigma456, line 2: Vscode debug supports AutoCAD 2021 and onward

Sen-real: This is a copy of existing statement here:
https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/137/files#diff-cb5c8ca542958828c6432e00b457edb26ac115c3f63e24d6cd54d98c548732dcR4

nigma456: Can we also update the readme.md as well?

LanDong-adsk (approved):

i18n/enu/out/debug.i18n.json
Sen-real: @nigma456 it's already the 2nd sentence in readme:
image

nigma456: Yes, I know. I mean AutoCAD 2021 and onward

PR #138, opened 2021-04-21
Part1 - Comment Command & Parsing

JD-Howard:
#### Objective
Added a right click menu "Generate Documentation" function that will detect the scope of the cursor and auto-generate something like a JSDOC header on user defined functions. 

This is part 1 of what will be 3 PR's on this branch. Trying to make it easily consumed by reviewers... The PR after this will be enhancing the auto-completion inputs and will be supporting these documentation headers in that feedback. The final PR will mostly a ton i18 placeholders for all the newly required localizations.

#### Abstractions
If the users current scope is inside a nested defun, then the user will be shown QuickPick options to select the defun they want to generate documentation for.

This PR does include Parser enhancements for interpreting this new documentation.

Also includes a new setting that will determine whether Mac OS users actually want to have their auto-completions filter out Windows-Only functions or not. I know if I was logged into a Mac and someone asked me to tweak some network script, I wouldn't want to switch machines or virtualize windows just to get access to something that could help me confidently make the changes and tell the user to try it again. Note, this PR doesn't really use this setting, but the next PR will be a lot to review so it made sense to include what I could.

Went ahead and migrated a few localization calls to the central one since I was in the file anyway. It clearly works the way it was, but the vscode-nls documentation says we shouldn't be recreating this.

#### Tests performed
I think the video speaks for itself, but note that a no-argument defun and no-defun scope at all scenarios were also tested. 

Also, added a test_case lisp file featuring these new comment blocks and an existing test file already dealing with the Parser/Namespace did have a test added for this new parsing context.

#### Additional
Forewarning, there may not be a lot left of the original completion lisp by the time its done, but the new version is being developed using a TDD approach so I will make sure it has excellent coverage. Primary reasons:
* General difficulty of bolting onto what was there and limitations that caused substantial bypassing. I did look at using a decorator proxy, but it just wasn't designed to support that.
* Adding flexibility for mac users does change a bunch of plumbing
* The very first function call did not detect all the things that could show up in a perfectly valid Lisp Atom
* Managing the users casing didn't appear to be necessary or even desirable now that we have the `webHelpAbstraction.json` file to lean on. For example, the Lisp documentation defines the case of `vlax-curve-getClosestPointToProjection`. The existing implementation would have inserted this as one of two extremely bad representations:
  * VLAX-CURVE-GETCLOSESTPOINTTOPROJECTION
  * vlax-curve-getclosestpointtoprojection

#### Screen shot
https://user-images.githubusercontent.com/51800232/115615667-1a07ff00-a2a4-11eb-86ca-3cba76dbb0ad.mp4


JD-Howard:
@Sen-real , I have some honey-do's to take care of, but later I will fire this up on a Mac to verify it all works there too.

Other than that this should be ready for review and is a precursor to other PR's I've already done substantial work on.

JD-Howard:
@Sen-real , the contents of this PR appear to be working fine on the Mac

@nigma456 , when I ran the tests `Formatter.test.ts` failed on 5, 13, 16 & 19, but only on Mac, they worked fine on my windows machine. I did run the maccompile.sh & npminstall, should be all I need to test those right? Also note the code coverage report is only showing `excludefile.ts` now.

hualin-whl:
This is very COOL feature!

extension/src/commands.ts
```diff
@@ -16,7 +16,7 @@ export function registerCommands(context: vscode.ExtensionContext){
 		}
 		catch (err) {
 			if (err){
-				let msg = localize("autolispext.help.commands.openWebHelp", "Failed to load the webHelpAbstraction.json file");
+				let msg = AutoLispExt.localize("autolispext.help.commands.openWebHelp", "Failed to load the webHelpAbstraction.json file");
```
hualin-whl, line 19: It (and othere similar changes in this PR) will break the Localization, see commit: https://github.com/Autodesk-AutoCAD/AutoLispExt/commit/fc630f78d3afefbf7842f1e596e1260e8f48af12

extension/src/commands.ts
```diff
@@ -32,12 +32,63 @@ export function registerCommands(context: vscode.ExtensionContext){
...
 				showErrorMessage(msg, err);
 			}
 		}
 	}));
 
 
+	// Associated with the right click "Generate Documentation" menu item
+	context.subscriptions.push(vscode.commands.registerCommand('autolisp.generateDocumentation', async () => {
+		try {
+			const pos = vscode.window.activeTextEditor.selection.start;
```
Sen-real, line 45: I suggest that we create a separate .ts file to hold the implementation, and call that implementation here

extension/src/format/parser.ts
```diff
@@ -364,3 +364,94 @@ export namespace LispParser {
         return result;
     }
 }
+
+
+export namespace LispParser {
+
+    interface ILspDocPair {
```
Sen-real, line 371: can we make a subfolder somewhere, to hold this class, plus [the code started here](https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/138/files#diff-30e8b6186a0e4537bb1f3855a033fc93fbbd1ed02b5907d99ce9a25a6525b975R45)?
it's not format anymore.

extension/src/resources.ts
```diff
@@ -95,6 +95,20 @@ export function getExtensionSettingString(settingName: string): string {
     return setting.toString().trim();
 }
 
+export function getExtensionSettingBoolean(settingName: string): boolean {
```
Sen-real, line 98: is this function used anywhere?

package.json
```diff
@@ -233,6 +238,11 @@
 					],
 					"default": "2022",
 					"description": "%autolispext.configuration.helptargetyear%"
+				},
+				"autolispext.completion.FilterMacOS": {
```
Sen-real, line 242: what is this fore?

extension/src/commands.ts
```diff
@@ -32,12 +32,63 @@ export function registerCommands(context: vscode.ExtensionContext){
...
+			args.forEach(a => {
+				insert += `  @Param ${a.symbol} $` + `{${count++}:?}${lf}`;
+			});
+			insert += '  @Returns ${' + `${count}:type?}${lf}|;${lf}`;
+			const snip = new vscode.SnippetString(insert);
+			vscode.window.activeTextEditor.insertSnippet(snip, defPos);
+		}
+		catch (err) {
+			if (err) {
+				let msg = AutoLispExt.localize("autolispext.help.commands.generateDocumentation", "A valid defun name could not be located");
```
Sen-real, line 87: @hualin-whl 
shall we add an item for this string in [the_owner_file].i18n.json by hand? 
e.g. if this code is kept here in this file, we then need change i18n\enu\out\commands.i18n.json. right?

extension/src/commands.ts
JD-Howard: Interesting I never saw this, I'll go read the documentation on vscode-nls again, I must have read it wrong or don't understand the scope of what I did read about not recreating this object.

extension/src/commands.ts
JD-Howard: Yes, I have been wanting to do that and this being something of an intermediate PR to break up some larger objective would make it a good one to take care of that.

extension/src/format/parser.ts
JD-Howard: Yes, I have been wanting to do that and this being something of an intermediate PR to break up some larger objective would make it a good one to take care of that.

extension/src/resources.ts
JD-Howard: Not right now, but it will be in the next PR. Here is the snippet of the next PR's information harvesting and its call to the decision maker.
```
const macFilter = AutoLispExt.Resources.getExtensionSettingBoolean('completion.FilterMacOS');
const acadVer = Number.parseInt(AutoLispExt.Resources.getExtensionSettingString('help.TargetYear'));
const isDCL = AutoLispExt.Documents.getSelectorType(document.fileName) === AutoLispExt.Selectors.dcl;
const winOkay = includeWindowsFunctions(os.platform(), isDCL, macFilter, acadVer);
``` typescript
Now that we have a release year targeting from my last PR, Mac users can now be included/excluded DCL functions in addition to VLA. So, I am layering in a lot more decision making under that checkbox and put 100% of that decision making into a single function taking a bunch of arguments so it can be easily tested.

File: package.json (hunk 5 above)
Comment (JD-Howard) [reply package.json:242]: This is from the abstraction section above:
<q>Also includes a new setting that will determine whether Mac OS users actually want to have their auto-completions filter out Windows-Only functions or not. I know if I was logged into a Mac and someone asked me to tweak some network script, I wouldn't want to switch machines or virtualize windows just to get access to something that could help me confidently make the changes and tell the user to try it again. Note, this PR doesn't really use this setting, but the next PR will be a lot to review so it made sense to include what I could</q>

File: extension/src/commands.ts (hunk 6 above)
Comment (JD-Howard) [reply extension/src/commands.ts:87]: In the Objective section, I suggested we get square with i18 in a final (or 3rd) PR just because it will muddy this up quite a bit and make it harder to sift through what needs reviewing. I will do it all up front if needed, just trying to be kind 😃

File: extension/src/commands.ts (hunk 1 above)
Comment (JD-Howard) [reply extension/src/commands.ts:19]: I think the actual problem has to do with our gulp file using `vscode-nls-dev` and that code using this regex
   `/^\s*(["'])vscode-nls\1\s*$/;` 
Which kicks out any file that doesn't directly import the `vscode-nls` reference.

I maintain that we have calls to `nls.config()` all over our code and according to everything I've read, this should not be happening. Per the documentation,  the `extension.ts` should own defining the `config()`

All other files should be importing `nls-config` because of the `vscode-nls-dev` anomaly and be creating their own localize reference using `loadMessageBundle()`

The `localize()` within the *context.ts* is entirely useless because of the limitation imposed by vscode-nls-dev, should be removed and purged from all existing documents; which I will add to this PR.

File: extension/src/commands.ts (hunk 2 above)
Comment (JD-Howard) [reply extension/src/commands.ts:47]: This was not implemented in the latest commit because I am not sure where to offload it to. Either the `commandTasks.ts` I previously suggested or maybe the `/src/help/` since this is kind of augmenting the data related to it?

File: extension/src/commands.ts (hunk 1 above)
Comment (JD-Howard) [reply extension/src/commands.ts]: FYI, purged `AutoLispExt.localize()` from all the drawings and used the `nls.loadMessageBundle()` to populate our `const` in those files. I tested the project commands responses previously using the AutoLispExt accessor and with Spanish; they popped up in Spanish as expected using a fresh VSIX compile. Can we assume they are functioning properly with `loadMessageBundle` based on that?

File: extension/src/format/parser.ts (hunk 3 above)
Comment (JD-Howard) [reply extension/src/format/parser.ts]: Parsing now has its own folder and everything parsing related is now in /src/parsing/

File: extension/src/commands.ts (hunk 6 above)
Comment (JD-Howard) [reply extension/src/commands.ts]: I added them, but of course that and your other comments to break stuff up has now blown up the number of changed files. I was trying to avoid that and do it incrementally, but its there now.

File: extension/src/commands.ts (hunk 2 above)
Comment (Sen-real) [reply extension/src/commands.ts:47]: good idea

File: i18n/chs/out/commands.i18n.json [hunk 7]
```diff
@@ -1,4 +1,5 @@
 {
   "autolispext.help.commands.openWebHelp": "无法加载 webHelpAbstraction.json 文件",
```
Sen-real, line 2: @hualin-whl 
I remember that we only need to add new resource string for enu, and localization team will update all the files for other languages, am I right?

JD-Howard: For the i18 folder, your question to Hualin is exactly how it was described to me. I just made that app to populate the empty entries for the other languages so it was more clear what they needed to back fill.

For the root language files, the gulp  process automatically adds the missing entries referenced in the package.json, but it propagates the available version (english) into all the files. Which is a little less intuitive than empty string, but should still be vividly clear what needs back filled.

extension/src/commands.ts
JD-Howard: Well that wasn't very clear so here are 2 options:
* Option1: `/src/commandTasks.ts`
* Option2: `/src/help/userDocumentation.ts`

Sen-real: I personally like opt. 2, but it's up to you.

i18n/chs/out/commands.i18n.json
Sen-real: Hi Josh, I checked our localization process. We (feature developers) only need to update files for "enu". We can save the effort here.

extension/src/commands.ts
Sen-real: Yes I built your code and tested against some of them. It looks properly translated.

i18n/chs/out/commands.i18n.json
hualin-whl: I agree here! Sorrry if I introduced some confusion here!
So for the pre-release extension we only support the ENU version. This is a clean and neat desicion.

extension/src/commands.ts
```diff
@@ -39,5 +41,56 @@ export function registerCommands(context: vscode.ExtensionContext){
...
 
 
+	// Associated with the right click "Generate Documentation" menu item
+	context.subscriptions.push(vscode.commands.registerCommand('autolisp.generateDocumentation', async () => {
+		try {
+			const pos = vscode.window.activeTextEditor.selection.start;
+			const doc = AutoLispExt.Documents.getDocument(vscode.window.activeTextEditor.document);
+			const exp = doc.atomsForest.find(p => p.contains(pos));
+
+			// Locate all the Defun & Defun-Q statements or an empty array if no actionable scope was found
```
hualin-whl, line 51: One comment code block is one "small" function, maybe we can refractor it to more readable.

JD-Howard:
@Sen-real , The bulk of the additions from this PR are in the 90%+ on coverage and its passing the checks. I think this is as good as its getting for the time being. Again, there will be more PR's.

JD-Howard:
Nothing I did should have caused that check to fail, pretty sure that is another glitch that just needs to be re-ran.

Sen-real:
> Nothing I did should have caused that check to fail, pretty sure that is another glitch that just needs to be re-ran.

It's random, I re-triggered that job and it passed.

Sen-real (approved):

PR #139, opened 2021-04-28
Fix ACD-19624: we should not append extra whitespace before right paren at the end of line

Sen-real:
##### Objective
Fix ACD-19624: we should not append extra whitespace before right paren at the end of line

##### Abstractions
As shown in the diff of formatedBasefile5.lsp in this PR, previously, when formatting the test file, an extra whitespace was noticed before the right paren at the end of current line:
image

This is because for all atoms of a given line - except the last atom, a whitespace was appended by [the code here](https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/d0182cb9555f7b2afe5e453b8e3411aa213f3938/extension/src/format/sexpression.ts#L772).

The fix is to call trimRight(), as lone as the ending atom of current line is a closing paren.

##### Tests performed
The bug is gone.
npm run test is green.

### DIFF
--- a/extension/src/format/sexpression.ts
+++ b/extension/src/format/sexpression.ts
@@ -822,8 +822,12 @@ export class Sexpression extends LispAtom {
         }
 
         // Last atom may be )
-        if (hasCloseParen)
-            res += this.formatLastAtom(startColumn, columnWidth, this.isMultilineString(res));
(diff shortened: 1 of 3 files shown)


LanDong-adsk (approved):

nigma456:
Good finding.

PR #141, opened 2021-05-06
Correct the usage of getConfiguration('autolispext') by modify the extensionDevelopmentPath

nigma456:
##### Objective
Update to use the correct extensionDevelopmentPath and remove mock method
##### Abstractions
In our lisp extension test we have wrongly set the extensionDevelopmentPath in runTest.ts so that it will fail to read the configuration('autolispext') and we have to use a workround of mock method for formatter.test.ts

In this subimission I corrected the extensionDevelopmentPath and removed the mock method

##### Tests performed
Passed local and CI test


.vscode/launch.json
```diff
@@ -22,6 +22,7 @@
             "preLaunchTask": "buildproject",
             "runtimeExecutable": "${execPath}",
             "args": [
+                "--disable-extensions",
```
JD-Howard, line 25: This was handled in PR #160 and quite a bit more. Make sure you update to the latest before these test additions get added.

extension/src/test/runTest.ts
```diff
@@ -6,7 +6,8 @@ async function main() {
 	try {
 		// The folder containing the Extension Manifest package.json
 		// Passed to `--extensionDevelopmentPath`
-		const extensionDevelopmentPath = path.resolve(__dirname, '../../../../');
+		// The __dirname is  \AutoLispExt\out\test
+		const extensionDevelopmentPath = path.resolve(__dirname, '../../');
```
JD-Howard, line 10: This was handled in PR #160 and quite a bit more. Make sure you update to the latest before these test additions get added.

PR #151, opened 2021-05-20
Update the change log for AutolispExt v1.5.0

Sen-real:
##### Objective
Update the change log for AutolispExt v1.5.0

##### Abstractions
The change log is updated to describe what's new since ver1.4.0.
Also updated i18n/enu/package.i18n.json with package.nls.json.

NOTE: extension version is not set to 1.5.0 yet.


ambrosl-adsk:
The only comment I have is related to this text "Determines the targeted AutoCAD year version of the 'Open Online Help' command." in _package.i18n.json_.

I would change it to "Specifies the targeted AutoCAD release of the 'Open Online Help' command." if the user can change it.

ambrosl-adsk:
Version in the package.json file reflects 1.4.2... should this be 1.5.0?  If not, there seems to be a discrepancy between the extension's version and change log.

Sen-real:
> Version in the package.json file reflects 1.4.2... should this be 1.5.0? If not, there seems to be a discrepancy between the extension's version and change log.

Sure, it will be changed to 1.5.0 right before we make a package to publish.

hualin-whl:
Happy to see we will release another new version!

PR #152, opened 2021-05-27
Update resource strings translated by localization team

Sen-real:
##### Objective
Update resource strings translated by localization team.

##### Abstractions
It's everything we should translate for 1.5.0.

##### Tests performed
image
image


i18n/enu/out/project/saveProject.i18n.json
```diff
@@ -1,4 +1,4 @@
 {
-  "autolispext.project.saveproject.noprojecttosave": "No project to save.",
-  "autolispext.project.saveproject.generateprjcontentfailed": "Failed to generate project content."
+	"autolispext.project.saveproject.noprojecttosave": "No project to save.",
```
LanDong-adsk, line 2: are these changes not necessary?

Sen-real: > are these changes not necessary?

It is necessary, because once we replace it with the entire file from localization team, we won't have t this diff anymore in the future.

Sen-real: You can turn on the following option to ignore them:
image

LanDong-adsk (approved):

PR #156, opened 2021-05-30
Update autocompletionProvider.ts

aeo000000:
##### Objective

1. Show help description in pop menu of suggestions.( the function name 'vla-set-XXX' must change to 'vla-put-XXX'  in webHelpAbstraction.json，its bug)

2. VScode has some issues to check the "word" range, exp: type "vl-str" and type esc or backspace,  continue to type "ing" and select 'vl-string->list' from menu,it will be: 'vl-vl-string->list' . 
Improved this problem, although it has not been completely solved, for example, typing ‘’-’’ still cannot be displayed as a complete word

##### Screen shot

image

### DIFF
--- a/extension/src/completion/autocompletionProvider.ts
+++ b/extension/src/completion/autocompletionProvider.ts
@@ -3,6 +3,9 @@ import * as os from 'os';
 import { isCursorInDoubleQuoteExpr } from "../format/autoIndent";
 import { allCmdsAndSysvars, internalDclKeys, internalLispFuncs, winOnlyListFuncPrefix  } from "../resources";
 
+//I haven't learned ts, I can only use js to modify it and hope it's useful。
+// the function name 'vla-set-XXX' must change to 'vla-put-XXX'  in webHelpAbstraction.json
+const InternalFuncHelpObj=require('../help/webHelpAbstraction.json');
 
 export function isInternalAutoLispOp(item: string): boolean {
     if (!item)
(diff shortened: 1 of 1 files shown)


LanDong-adsk:
Hello @aeo000000 , before we can merge your PR, please sign this file and send to me (lan.dong@autodesk.com). thanks!
https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/main/Autodesk_Form_Ind_Contrib_Agmt_for_Open_%20Source.pdf

JD-Howard:
@aeo000000 Bullet #1 really needs to be fixed in the C#/JSON. Treating the symptom isn't really a solution.

I really should have noticed this, but think the biggest reason I missed it has to do with your Bullet #2. There are all kinds of things I use in lisp (like colons) that yield nothing worth using so I don't even pay attention to intellisense involving any special characters.

Anyway, unless you really want to dig into a complete nightmare, I'll go fix Bullet #1 in the C#/JSON. I actually plan to rewrite that "prototype" producing the JSON to be more sustainable and with better technologies sometime in the coming months.

extension/src/completion/autocompletionProvider.ts
```diff
@@ -76,28 +85,29 @@ function getCompletionCandidates(allCandiates: string[], word: string, userInput
...
-            word = linetext.substr(pos + 1, length);
-            break;
-        }
-        length++;
-        hasSetLen = true;
-    }
+//     let pos = position.character;
+//     pos -= 2;
+//     let length = 1;
+//     let hasSetLen = false;
```
hualin-whl, line 99: If you don't need the codes, please remove it

PR #157, opened 2021-06-01
Revert PR 150

Sen-real:
Reverts Autodesk-AutoCAD/AutoLispExt#150.

With https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/150, in order to release v1.5.0, an incomplete work was temporarily reverted.

Since we have now published v1.5.0 to the market place, the reverted code can be added back.


Sen-real:
@JD-Howard
should we revert this PR, if we don't have time to implement this option right now?

PR #160, opened 2021-07-13
RenameProvider & a lot of supporting infrastructure

JD-Howard:
### Objective
Introduced 'F2' renaming functionality. Lisp not having any explicit means of contextually importing functions/variables made any solution sketchy. After a lot of considerations the "@Global" comment documentation keyword is the official "trigger" for what will be a valid cross-document renaming target. With that said, only 1 instance (of @Global) has to exist anywhere in the workspace, project or opened documents to make the keyword applicable across all of them. Do note, any instance using it will be "verified" that the tagged keyword wasn't localized; thus canceling the global flag. Finally, this is all perfectly contextual, a series of nested, but identical variable names are all scoped and acted upon according to their scope.

I will add that this infrastructure supporting F2 was also extremely foundational (new tooling) for cleaning up and/or enhancing Goto, AutoComplete and signature assistance.

### Abstractions
Performance was a major obstacle in this PR. The LispContainer parser was reverted to the original non-index targeting version, then it was optimized to death until it was a full 2x faster than the formatting parser. Next I layered on additional value enhancements such as linear indexing, comment linking, non-native symbol aggregation; which caused another round of scavenging for performance. Leading to a StringBuilder class that leaned on character codes instead of character x character assembly of strings and that finally hit a sweet spot of features vs performance.

One important concept introduced in this PR is a "Flat View" of the document tree. Through extensive testing, it was proved to be an exceptionally performant process to flatten a LispContainer. With the current version of the parser pre-tagging every single LispAtom with its flattened index, linear traversal for things such as extracting  "contextual" comments becomes very performant. These flat indexes were also used to store primitive pointers to concrete objects; which has all kinds of memory, garbage collection and performance benefits. These will be the cornerstone of obtaining workspace autocompletion/signature data without major impacts to the user experience.

Configuration was another big change. Our "runTests.ts" was targeting 2 directories back further than it should have, which prevented the extension from loading, which oddly created more results than targeting the right directory. I eventually figured out this was tied to my ongoing issues with command line tests/coverage. Basically, they were running, but with a pre-opened drawing causing a scope conflict with the extension activation. There are now explicit operations in the tests to load a workspace, activate the extension, open files, and most importantly force close them before exit. It now doesn't matter if you run tests/coverage from vscode or from a command line, they should all "act" exactly the same way now...

A lot of things were touched, but I do think I've gone over everything enough times it should be a slam dunk if CI/CD doesn't baulk at my configuration changes.

### Tests performed

There is definitively nothing part of this PR that isn't 95%-100% on test code coverage. In fact, the configuration work performed drastically improved the overall coverage of the root 'src' file groups.

Edit: the image below shows `/src/analyze/` but that was actually renamed to `/src/services/` just before the PR went up.

src aggregate

src compare

### Screen shot
Here is a demo of the new feature in action

https://user-images.githubusercontent.com/51800232/125433539-d5c72753-b415-486d-9a1e-4f7f035f04f3.mp4

### DIFF
--- a/extension/src/documents.ts
+++ b/extension/src/documents.ts
@@ -5,12 +5,15 @@ import { AutoLispExt } from './extension';
 import { ProjectTreeProvider  } from "./project/projectTree";
 import * as fs from	'fs-extra';
 import { glob } from 'glob';
+import { DocumentServices } from './services/documentServices';
+import { SymbolManager } from './symbols';
 
 
 enum Origins {
 	OPENED,
 	WSPACE,
-	PROJECT
+	PROJECT,
+	UNKNOWN
 }
 
 
@@ -132,7 +135,13 @@ export class DocumentManager{
 
 	private tryUpdateInternal(sources: DocumentSources){
 		if (sources.native && (!sources.internal || !sources.internal.equal(sources.native))) {
-			sources.internal = ReadonlyDocument.getMemoryDocument(sources.native);
+			sources.internal = ReadonlyDocument.getMemoryDocument(sources.native);			
+			if (DocumentServices.hasUnverifiedGlobalizers(sources.internal)) {
+				// symbol mapping actually takes slightly more time than parsing, so the goal is
+				// to keep a persistent representation of anything containing @global exported
+				// variable/function name. Use cases are Rename, Autocomplete and Signature helpers
+				SymbolManager.updateOrCreateSymbolMap(sources.internal, true);
+			}
 		}
 	}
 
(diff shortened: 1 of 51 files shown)


JD-Howard:
@Sen-real & @hualin-whl all checks have passed...
This is a large PR, but it is all directly related to the same difficult task... I have not tested this on my Mac yet, but I will fire it up tomorrow, verify all expected tests still pass, and behaves as expected

extension/src/documents.ts
```diff
@@ -176,7 +185,17 @@ export class DocumentManager{
...
+	}
+	
+	tryGetDocument(fsPath: string): ReadonlyDocument {
+		// This is something of a hack to query an existing document from a randomly acquired file path.
+		// ultimately, if the LSP file exists, it will return a document, but it may or may not be part
+		// of the "normal sources" and thats why the origin is unknown. More often than not, this will
+		// yield what it already processed in a production setting. However, while running tests this
+		// UNKNOWN version could be the only origin and won't be useable in any "non-testing operations"
+		const key = this.pathConsumeOrValidate(fsPath, Origins.UNKNOWN);
+		return this._cached.get(this.normalizePath(fsPath))?.internal;
```
JD-Howard, line 197: Yeah I should have used the key instead of re-normalizing the fspath

JD-Howard:
@Sen-real , I had to update all the path normalizer(s) to favor forward slashes instead of backslashes and addressed all the CodeQL annotations. All of my stuff is passing just fine on the Mac and I did manual testing too. 
However, the same 4 LispFormatter tests that have always been a problem on Mac are still failing, but I am not going to dig into why; at least not in this PR.

i18n/chs/out/commands.i18n.json
```diff
@@ -1,5 +1,6 @@
 {
   "autolispext.help.commands.openWebHelp": "无法加载 webHelpAbstraction.json 文件",
   "autolispext.commands.addFoldingRegion": "无法插入代码段",
-  "autolispext.help.commands.generateDocumentation": "找不到有效的 defun 名称"
+  "autolispext.help.commands.generateDocumentation": "找不到有效的 defun 名称",
```
Sen-real, line 4: We don't need to do it on our own. Before publishing a new release, we'll send English resource files to the globalization team, and they'll provide us with all the non-English resource files.

Sen-real (approved):
Thanks a lot for this great new feature! please let us know when it's ready to merge.

Issue #88, opened 2021-02-23
Go To Defnitition inside if goes to wrong statement


hansWurst-creator:
**Describe the bug**
Go To Defnitition inside if jumps to wrong statement

Test with the following code:
`(defun foo () 
  (setq a 1)
  (if T 
    (progn 
      (setq a 2)
      (print a)
    )
  )
)`

- set the cursor on a in `(print a)`
- Press F12
- cursor cumps to `(setq a 1)`

![image](https://user-images.githubusercontent.com/70693364/108841445-0c9e0380-75d8-11eb-8047-45d7b4d6ff85.png)

**Expected behavior**
Courser jumps to (setq a 2)

hualin-adsk:
I think maybe we have some mis-understanding here, for these two` (setq)` line codes, because there is no local function variable definition like:
```
  (defun foo (/a)

)
```

So for this case the `a` is a "**global**" variable, so it goes to the first` (setq)` codes by intent as we define the behavior initially.

hansWurst-creator:
So then Go to Definition will only be really helpful for functions and not variables?
Because with local variables you always end up in the defun statment.
For proper use with local variable the Go to Implementation would be the correct choice?

hualin-adsk:
We want to the F12 can support it for Global variables, and I think now it works well for the Local variables.
I am trying understand what is the "correct" behavior for this global variables case mentioned above.

hansWurst-creator:
So with localized variable this is the current implementation:
![image](https://user-images.githubusercontent.com/70693364/108864096-2a786200-75f2-11eb-86e2-c997677cac66.png)

This looks inconsistent with how local functions are handled:
![image](https://user-images.githubusercontent.com/70693364/108864634-abcff480-75f2-11eb-9050-b4899ee15cba.png)

I was expecting a similar behavior with value variables. More like when was the last time something assigned to a certain variable not that much in what scope it lives in.

Well Lisp is quite a different programming language from statically typed languages. So it is not easy to bring over established principles to Lisp. I really appreciate the work though. Even the feature in its current state is a huge step forward.

JD-Howard:
It kind of looks like globalized functions are just finding the first declaration within its active function scope. Which would make its behavior pretty consistent with a localized variables behavior. 

[...]

PR #167, opened 2021-09-21
Provider Enhancements

### LINKED ISSUE #154: Fail to go to definition on Mac unless open the lsp file
**Describe the bug**
Mac Only. Fail to go to definition unless open the lsp file with definition of function. 
**To Reproduce**
1.  Open folder in vscode. The folder includes a/a.lsp and b/b.lsp.
2. a.lsp defines function foo(), b.lsp invokes function (foo)
3. Ensure a.lsp is not opened in vscode

**Expected behavior**
(foo) in b.lsp could goto (defun foo) in a.lsp

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Mac OS X
 - Version 1.4

**Additional context**

JD-Howard:
### Objective
This is mostly an update of `DefinitionProvider` to use the infrastructure put in place by the `RenameProvider`, but it also adds a `ReferenceProvider` that will essentially provide some visibility into the scope of the `RenameProvider`.

### Abstractions
The DefinitionProvider was previously discussed to be a little squirrely on issue #88. Also hoping this will resolve #154 since I did test the F12 functionality on my Mac. The new "behavior" of the F12 now supports the `@Global` comment tags as its highest priority; which means the new version can support global variables from other documents that the original DefinitionProvider intentionally omitted. This was made possible by the infrastructure added in the RenameProvider PR. Also note that the new DefinitionProvider is very focused on finding the definition, the old one made attempts to "walk" up a document, but now that only happens if it is a non `@Global` & non localized variable reference. IE, when there is no clear source.

#### General Changes:
- Complete rewrite of the original DefinitionProvider code and removed some of its associated support functions as well.
- Removed the ProviderShared tests since they were all related to the old definition provider.
- Minor maintenance to *symbols.ts* to rely more on interfaces and prevent parent/child constructors from ever propagating outside of the SymbolManager. Which also updated a few other random files to remove the concrete class references.
- Expanded the *flatContainerServices.ts* with new functions and optional argument context.
- Fixed a non-critical context bug I found in the *flatContainerServices.ts* affecting the RenameProvider.
- Changed our vscode target version in the *package.json* to 1.57 because our extension was no longer supporting the 1.50 we had it set to in all contexts. I also updated the *runTests.ts* to read & use the version we are specifying in the *package.json* so we can hopefully avoid falling behind on this in the future.

  

### Tests performed
I believe everything related to this PR has at least 97% test coverage with the exception of what was added to the *commands.ts* since I can't think of a good way to test those entry points. However, they shouldn't really matter since they defer almost everything to the testable code.
- Overhauled: providers.gotoProvider.test.ts
- Deleted: providers.providerShared.test.ts
- Added: providers.referenceProvider.test.ts
- Updated: services.flatContainerServices.test.ts

### Screen shot
The mock workspace I setup during the rename provider doesn't do anything to illustrate multiple potential references so I'll have to explain that behavior. The current F12 functionality will prefer an `@Global` marked function, but if there is no globally marked function and multiple identically named defun's exist, then it would present a list of possible definitions just like it did in the original implementation.

https://user-images.githubusercontent.com/51800232/134130230-09bc306b-9bdb-4eab-8253-d09f12d9e893.mp4


JD-Howard:
@Sen-real 
Sorry this took so long to get out. Most of it was done a while ago, but my CompTIA Network+ turned into a nightmare of acronyms and magic numbers that required all of my attention. Glad I did it, but quite the memory exercise...

Also note, I need to investigate the change associated with 500dcb77ed3af0fc2764ad4746fe9c4f49b21292 and see if caused any issues. However, this is an issue related to our target version being 1.50 and me changing it to 1.57. So, if there is a problem, then it was already there and we just weren't aware of it. I assume, simply because the AutoCompletion has pretty good test coverage these days, but I'll check on it tomorrow.

JD-Howard:
I have confirmed that 500dcb7 did not cause any issues. VSCode added an alternate data type to their extension, but we are assembling the concrete class we use with a string. So, the added `toString()` was just a formality to downgrade the new interface feature to the primitive we were already providing.

I also confirmed #154 will be resolved by this PR.

Sen-real:
Great work, thanks!

Sen-real (approved):

PR #170, opened 2021-09-28
Maintenance

JD-Howard:
### Objective
This is purely a cleanup and maintenance PR. We've had *odd locations* for some core objects and some legacy redirection for a while and it was time to get that sorted out.

### Abstractions
All refactoring was performed using WebStorm for maximum accuracy and it was put through manual and automated testing after the refactoring. I chose to put our core objects representing lisp syntax objects in the "astObjects" folder because they are fundamentally "Abstract Syntax Tree" representations and that is the common term used by other language parsing frameworks such as Esprima. Here is the sum of what happened:

- Deleted the *parser.ts* from a previous refactoring and updated any references that had remaining legacy pointers
- Moved *ILispFragment* to its own file in the *astObjects* folder
- Moved *LispAtom* to its own file in the *astObjects* folder
- Moved *LispContainer* to its own file in the *astObjects* folder
- Moved *Sexpression* to its own file in the *astObjects* folder
- Deleted the `getRawText` function from the `LispContainer` because nothing was using it and the code coverage went up to 99% with it gone. The `AsText` function also does something similar and is considered more complete for use in verifying zero loss (except some spacing) during the parse.

Other than what was noted in that list, there is no functionality changes in this PR.

### Tests performed
All 202 tests are passing. Also checked the code coverage.


Sen-real:
@JD-Howard thanks a lot for the cleanup! Is it ready for review?

PR #202, opened 2022-09-06
Hover Provider

JD-Howard:
### Objective
I originally expected to finally address the long standing issues with AutoCompletion, but one of the prerequisites for that is providing "details" about the suggestions. So, I finished a markdown enhanced **HoverProvider** and the completion updates will have to be in another PR; probably for the best since this wound up having a lot of maintenance updates too.

### Abstractions
The presentation layer is self contained and works for native AutoLisp/DclTiles/DclAttribute identifiers if they exist in `webHelpAbstration.json` or defined by the users code. This uses previous developments to get definitions from other files, but if it is used and not defined in the current file, then it does require the user to have an `@Global` comment to be applicable; examples of pass & fail situations provided in the video. Also note, we are a bit behind on the latest VSCode versions, the newer releases have HTML capabilities and we are restricted to Markdown right now. I don't think this is an issue, but I never got the local `vscode://filePath:line#` links working. So, kind of wondering if they would using newer references. For the time being, I just added a "source" to the popup to give a clue about the origin.

Uses previous developments for generating/parsing user provided documentation to enhance user defined function presentation.

There is 1 new localization, but I didn't put all the placeholders in i18n files.

Generally not married to the current presentation, but it will "do for now" until `webHelpAbstraction.json` gets updated. There is a test that will regenerate all the baseline MDs given a specific boolean flag. It is the obligation of whoever tinkers with the presentation code to regenerate them, manually review the outputs for "expected" results and include them in a PR to keep the tests passing. I left a similar note next to the variable.

Markdown generation does account for argument index emphasis. This will be useful for a future signature provider.

This addition makes me consider deleting the context menu "open web help" functionality in the future. 

Split the original web help file into a Library & Object files. Very little changed even if it all looks like new code. Its safe to say the Library was the only thing that changed in any meaningful way.

WebStorm was used for a lot of the major refactoring. 

### Tests performed
The "new" content in this has ~95% code coverage, but would like to do some further testing on "User Defined" representations through mocks at some point.

All existing tests are still passing, but note that I did break a number of them during some cleanup. When your looking at test changes that aren't specifically in `providers.hoverProvider.test.ts` just keep in mind:
- that some redundancy was removed that only tests were using
- DocumentManager used to be exclusively for LSP files and now holds DCLs too.
- Misc core things changed their access points to fix a load priority problem; IE various files were loading each other and not necessarily in an order that hooked everything up at the time of the extension entry point.

### Screen shot
It is kind of hard to see my mouse in this, but I am in fact hovering over the various things that you see pop up and they often highlight the atom shortly after.

https://user-images.githubusercontent.com/51800232/188553502-db1ae887-53f2-453d-b0ff-0db86fc0d86c.mp4


JD-Howard:
Regarding the failed tests, pretty sure this is a CRLF vs LF problem. I'll normalize the tests and try again.

I don't know what is up with your CLA email check, but I am not a fan of you trying to impose requirements that affect **MY** personally identifiable information. Putting my email address in public view is a hard no...

Sen-real:
> Regarding the failed tests, pretty sure this is a CRLF vs LF problem. I'll normalize the tests and try again.
> 
> I don't know what is up with your CLA email check, but I am not a fan of you trying to impose requirements that affect **MY** personally identifiable information. Putting my email address in public view is a hard no...

As the bot says, individual contributes could either send us signed CLA file or sign it online.
You have already sent us the file, so please just ignore that message.

JD-Howard:
That was a lot of tinkering to get those tests to pass. The CI/CD environment has a lot of things that don't work the same way as my local machine; LineEndings, readonly VSCode properties, readonly workspace settings.

Note that this has not been tested on a Mac yet, but I'll do that tomorrow.

Sen-real:
> Note that this has not been tested on a Mac yet, but I'll do that tomorrow. Also, not sure what is going on with CodeQL, I'll review and fix those if they ever show up. Could the CLA thing be preventing it from reporting?

CodeQL is temporarily turned off, and will be turned on after it works again.
I'll ask CLA team if it's allowed to add an option like "Don't ask me again".

JD-Howard:
Mac testing is completed, 1 bug in the test itself was found.

extension/src/resources.ts
```diff
@@ -95,6 +99,17 @@ export function getExtensionSettingString(settingName: string): string {
     return setting.toString().trim();
 }
 
+export function setExtensionSettingString(settingName: string, newValue: string): boolean {
```
Sen-real, line 102: is this used any where?

Sen-real (approved):

PR #203, opened 2022-09-28
DCL Auto-Completion Provider

JD-Howard:
### Objective
Add a highly assistive DCL Language AutoCompletionProvider with attached documentation about the suggestions.

### Abstractions
This is the first meaningful use of the new DCL parser and was designed to "assist" the user in ways that create clean looking DCL. Especially necessary given the lack of a DCL formatter. When a formatter is available, this should be revisited because CompletionItems have the ability to trigger a format command and that will help handle a few rare edge cases that may currently produce odd results. With that said, using this in the context of an existing DCL that has odd formatting or excessive indentation may have different ideas about where indentation belongs. Again, this is unavoidable without a formatter. However, if someone is building a DCL from scratch, this should be extremely useful. It filters out possibilities as they are used/burned it only shows attributes that are applicable to the active tile, and won't show attributes in situations that don't make sense.

Also note, some of the outstanding comments from PR #202 were addressed here.

### Tests performed
Still working on tests and this has not been tested on a Mac yet. Currently have an average of 87% code coverage.

### Screen shot

https://user-images.githubusercontent.com/51800232/192707497-eeb4cc32-6cc0-4c43-b0c6-5d16d1b19999.mp4


JD-Howard:
@Sen-real 
I'll write the `NoAtom, HasPosition with ProximityParent` tests and give it a spin on my make tomorrow. I suspect that will bring it to an average of 95% on the code coverage. The nature of this caused extensive manual testing because I can't trust VSCode will actually show the suggestions I provide it. With that said, I consider it will tested even though I have outstanding unit tests to make. Give it a once over, fire it up for yourself and if you don't see any particular problems, then I would say its probably fine to include in 1.6; Certainly better than the nothing we currently have for DCL.

Sen-real:
Awesome new feature, cool!

Sen-real (approved):

extension/src/completion/completionLibraryDcl.ts
```diff
@@ -0,0 +1,183 @@
...
+
+
+		this.dclSnippets.set(SnippetKeys.STRING,
+			this.makeSnippet('"?"', 'string structure', 'Primitive', '"$0"', Kinds.PRIMITIVE)
+		);		
+		this.dclSnippets.set(SnippetKeys.TRUE,
+			this.makeSnippet('True', 'boolean structure', 'Primitive', 'true', Kinds.PRIMITIVE)
+		);
+		this.dclSnippets.set(SnippetKeys.FALSE,
+			this.makeSnippet('False', 'boolean structure', 'Primitive', 'false', Kinds.PRIMITIVE)
```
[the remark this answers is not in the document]
JD-Howard: Ouch, it did not occur to me that this introduces a bunch of of potential localizations. 
Yeah, they probably should be localized

extension/src/completion/completionLibraryDcl.ts
```diff
@@ -0,0 +1,183 @@
...
+		snip.documentation = desc;
+		snip.sortText = label === 'dialog' ? '!!!dialog' : label;
+		snip.kind = kind;
+		return snip;
+	}
+
+
+	private generateDynamics() : void {
+		this.dclSnippets.set(SnippetKeys.DIALOG,
+			this.makeSnippet('dialog', 'Generates a dialog structure', 'Snippet', `\${1:NAME} : dialog {\n\t$0\n}`, Kinds.SNIPPET)
```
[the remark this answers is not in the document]
JD-Howard: I think the description is fine in the context of the english language. This is a drastically enhanced completion provider so a great deal of this contextually inserts snippets and it isn't the intent to be presented as a snippet specifically. IE, the context floats around a lot. So, maybe I should just change all the `Kinds.SNIPPET` to be structure and make it match everything else. I'll just have to do a sanity check on whether that will screw up any conditions.

extension/src/test/suite/completionItemDcl.test.ts
```diff
@@ -0,0 +1,58 @@
...
+const symbolsFileTest = path.resolve(extRootPath, "./extension/src/test/SourceFile/test_case/symbols.lsp");
+const commentsFileTest = path.resolve(extRootPath, "./extension/src/test/SourceFile/test_case/comments.lsp");
+//const markupsFileTest = path.resolve(extRootPath, "./extension/src/test/SourceFile/test_case/pdfMarkups.lsp");
+const largeFileTest = path.resolve(extRootPath, "./extension/src/test/SourceFile/unFormatted10.lsp");
+
+
+// NOTE: Just because these tests return results and pass, manual testing is still required. VSCode can be given
+//		 whatever we want and still choose to kick them out from the users perspective under various conditions.
+
+suite("Parsing: DocumentContainer Tests", function () {
```
JD-Howard, line 20: This was an early test fixture. The entire file needs to be deleted. The tests are all inside `provider.completionDcl.test.ts` now.

JD-Howard:
I added in the English localizations. but I made a judgement call on stuff like "Tile", "Attribute" and "Enum" being too native to DCL to mask or completely universal to all programming.
I have now tested what I do have on the Mac and it passed without issue.
I still have 1 test to build out and will do that later tonight.

i18n/enu/out/providers/hoverProvider.i18n.json
```diff
@@ -0,0 +1,3 @@
+{
+	"autolispext.hoverProvider.ambiguous": "Multiple definitions containing the @Global flag"
```
Sen-real, line 2: @JD-Howard 
When testing on this message I see this:

image

I guess the string` $(error) `at the beginning is not the expected result?

Reading the source code below I feel that an error icon is expected?
https://github.com/Autodesk-AutoCAD/AutoLispExt/pull/202/files#diff-e176c092ee19a3712e1d5938d6b89322fc76dee085e6cf2ccb55283e9fd6f2b7R154

.gitignore
```diff
@@ -1,5 +1,5 @@
-node_modules
-out
+/node_modules
+/out
```
Sen-real, line 2: the old `out` seems out now, have to add a `/` to make new added localization .JSON files visible to the `git diff` command

i18n/enu/out/completion/completionLibraryDcl.i18n.json
```diff
@@ -0,0 +1,7 @@
+{
+  "autolispext.commands.dclcompletion.primitive": "Primitive",
+  "autolispext.commands.dclcompletion.primitive.dialog": "Generates a dialog structure",
+  "autolispext.commands.dclcompletion.primitive.structure": "structural primitive",
+  "autolispext.commands.dclcompletion.primitive.string": "string structure",
+  "autolispext.commands.dclcompletion.primitive.boolean": "boolean structure"
+}
```
Sen-real, line 7: I moved them from `i18n/enu/out/commands.i18n.json` into this file, otherwise their localized versions can't be loaded.
The key is the `completion/completionLibraryDcl` in file path, which is identical to the .ts source file.

JD-Howard: Thanks for fixing that

JD-Howard:
The average code coverage on the DCL Completion is at 96%
I'd like it better if it had a formatter to smooth out a few edge cases, but I think this is probably ready to go.

Sen-real:
> The average code coverage on the DCL Completion is at 96% I'd like it better if it had a formatter to smooth out a few edge cases, but I think this is probably ready to go.

Thanks! it looks pretty cool.

PR #206, opened 2022-10-12
Make the localization identical on the DCL auto-completion UI

Sen-real:
##### Objective
Make the localization identical on the DCL auto-completion UI

##### Abstractions
It's noticed that the localization on the DCL auto-completion popup is not identical:
- sometimes it's fully localized
- sometimes its title is not localized
- sometimes it's not localized at all

This PR is to make the behavior identical - i.e. always fully localized.

##### Tests performed
The inconsistency found during testing is gone.


JD-Howard (approved):
Sorry I made an assumption about those localization being agnostic. These updates good to me.

PR #228, opened 2023-03-31
Documentation updates related to acet-* functions

ambrosl-adsk:
##### Objective
Provide access to the online documentation for the _acet-*_ functions that are documented.

##### Abstractions

- Updated the _webHelpAbstraction.json_ in the _ExtractHelp.zip_ file
- Updated the **Help: Target Year** setting to include the **2024** value and set it as the default value

##### Tests performed
Added all of the documented _acet-*_ functions to an LSP file and made sure each opened the correct online help topic with the "Open Online Help" command.  I also checked the syntax that comes up in the tooltip when hovering over each function.

##### Screen shot
image
image
image


JD-Howard:
@ambrosl-adsk 
Could you please replace the [webHelpAbstraction.json](https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/main/extension/src/help/webHelpAbstraction.json) instead of updating the zip file.

The zip really exists as an artifact from the C# processing. Since this is all pencil whipped, there really isn't a reason to update the zip.

ambrosl-adsk:
I have read the CLA Document and I hereby sign the CLA

ambrosl-adsk:
> @ambrosl-adsk Could you please replace the [webHelpAbstraction.json](https://github.com/Autodesk-AutoCAD/AutoLispExt/blob/main/extension/src/help/webHelpAbstraction.json) instead of updating the zip file.
> 
> The zip really exists as an artifact from the C# processing. Since this is all pencil whipped, there really isn't a reason to update the zip.

Sen pointed that out to me... and was updated separately in a new commit to this PR.

JD-Howard:
**acet-reg-prodkey** exists in the docs but not our highlighting syntax. We just need to add it to line 163 on `\extension\syntaxes\autolisp.tmLanguage.json`

I checked all the guid links but I didn't audit any of the actual documentation. I saw random things in there (missing ?) that isn't congruent but its also not important.

Generally this looks pretty good and more importantly functional. One thing you should probably consider doing though... 

#### Move `acet-ui-progress` into the `ambiguousFunctions`
Basically that function has multiple distinctly different signatures and thus really belongs in ambiguousFunctions. I also noticed that the official docs do not directly (in signatures) list the perfectly valid no-args arrangement.

ambrosl-adsk:
@JD-Howard `acet-reg-prodkey` is no longer supported, I did include it in the documentation/syntax as a way to indicate it is Obsolete.  So the short description does start with **(Obsolete)**.

I wasn't exactly sure about the **?**, some of the existing had it in places under syntax but not the argument.  Assuming its intent is to indicate optional arguments.  If the ? should follow the argument name in the JSON I can add it to be consistent.

`acet-ui-progress` without any arguments is really the syntax of `(acet-ui-progress [label [max]])`.  I can move it to the ambiguousFunctions where `vlax-3D-point` is located and parallel its signature.

JD-Howard:
> I wasn't exactly sure about the **?**, some of the existing had it in places under syntax but not the argument. Assuming its intent is to indicate optional arguments. If the ? should follow the argument name in the JSON I can add it to be consistent.

You generally the **?** correct in the signatures, they just often didn't exist in the arguments sections. Which is fine and not a big deal. The other thing I did notice is that the documentation had "named" versions for enums + the integer value. If the names exist (IE, are symbols that resolve to integers) then we should just use the names. Again, at this point I don't consider that much of an issue. The parser (when updated) can pluck those out later.

Sen-real (approved):
cool

Sen-real:
I have read the CLA Document and I hereby sign the CLA

JD-Howard:
@ambrosl-adsk 
I consider this temporary and I don't care either way, but Sen requested my review and I am just waiting to figure out if that 1 function is getting migrated into the ambiguous area. 

At minimum, `acet-reg-prodkey` may be deprecated, but it still valid for some environments and thus needs to be syntax highlighted.

Sen-real:
> At minimum, `acet-reg-prodkey` may be deprecated, but it still valid for some environments and thus needs to be syntax highlighted.

Yes I agree, as it already works well with online help, it should also be highlighted. I'll add it into the highlighting config file.
image

Issue #247, opened 2024-05-15
Intellisense :vlax bug


EnM4st3r:
**Describe the bug**
Somehow the suggestions for :vlax... are doubled. As you can see in the screenshot nearly every suggestion is doubled.

**To Reproduce**
1. Write an opening paranthesis
2. write ":vlax.."

**Expected behavior**
Only one suggestion

**Screenshots**
![image](https://github.com/Autodesk-AutoCAD/AutoLispExt/assets/163658977/09175b8b-92fe-469f-99a1-99950b04f4e0)

**Desktop (please complete the following information):**
 - OS: Windows 10

**Additional context**
Also when using these :vlax.. intellisense suggestions, it does not replace the already written colon but writes another one so when i use the autocompletion here it looks like this: ( ::vlax-true)
This kinda destroys the flow

Issue #249, opened 2024-06-10
Help: Target Year 2025


EnM4st3r:
Hi, can the Help Target Year for 2025 be added?

In settings:
![image](https://github.com/Autodesk-AutoCAD/AutoLispExt/assets/163658977/2a130811-2264-40b4-9bb5-edea4bc3bb2a)

Sen-real:
Thanks, @EnM4st3r , I'll make an update on this soon.
I'm also thinking if there's a better way that can avoid making an update every year on this field ...

PR #251, opened 2024-06-18
Fix issues #247 & #249

Sen-real:
##### Objective
Fix issue #247 and issue #249.

There's some unexpected problem with package-lock.json, so it's not updated in this PR and will be handled later.

##### Tests performed
Tested the 2 issues.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,7 +39,13 @@ jobs:
       - name: Build
         run: |
           python pack.py
-        
+
+      - name: Upload package-lock.json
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -39,7 +39,13 @@ jobs:
       - name: Build
         run: |
           python pack.py
-        
+
+      - name: Upload package-lock.json
+        uses: actions/upload-artifact@v2
+        with:
+          name: localFile
+          path: package-lock.json
```
Sen-real, line 47: will be reverted later.

PR #253, opened 2025-08-16
Adds Outline function list for LSP and DCL files.

JimsTools:
##### Objective
Renders the Outline showing "Executable Functions", "Functions" and "Test Functions" (those that begin with "#", not configurable, but probably should be). Also renders Outline for DCL files showing partial and complete dialog functions.

##### Abstractions
I am sorry, I don't know what this means.

##### Tests performed
Open LSP or DCL file.

##### Screen shot
image


JimsTools:
I have read the CLA Document and I hereby sign the CLA

qiuradsk:
Thank you for contributing to AutoLispExt!
I suggest re-using the existing infrastructure for extracting symbols instead of using Regex. The existing infrastructure is more efficient and can handle a wider range of symbols.

